PR #3, opened 2014-02-07
Ensure set_actor signal disconnection

vfigueiro:
Mainly added sender kwarg to pre_save.disconnect, because the signal wasn't getting disconnected for me without it.

Timestamped the request and used that to connect/disconnect the signal, since the request could be modified by another middleware, which would prevent correct match on disconnect.

Added process_exception too.


adamwojt:
For me the signal is disconnected too soon it seems. IT works when I override `__call__` and don't trigger process_response`. Could this be bug ?

PR #14, opened 2015-02-12
- Compatibility with django-polymorphic

johnrtipton:
django-polymorphic self._get_pk_value returns the class object.  If
this happens, we try adding _id to the pk_field.

### DIFF
--- a/src/auditlog/models.py
+++ b/src/auditlog/models.py
@@ -63,7 +63,12 @@ class LogEntryManager(models.Manager):
         Get the primary key field value for a model instance.
         """
         pk_field = instance._meta.pk.name
-        return getattr(instance, pk_field, None)
+        pk = getattr(instance, pk_field, None)
+
+        # Check to make sure that we got an pk not a class object.
+        if hasattr(pk, '__class__'):
+            pk = getattr(instance, '%s_id' % pk_field, None)
+        return pk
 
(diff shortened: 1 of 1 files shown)


jjkester:
Thanks for your contribution :-)

I have reviewed the code and it seems to be in order.

I am hoping to find some time to do some work on auditlog within the upcoming weeks and will merge your pull request then. (I am also planning to release a new version on PyPI).

jjkester:
I looked at your code again, and I have a suggestion to improve it a bit. At the moment your code checks for a ```_meta``` attribute in the ```pk``` value. A better way to formulate this condition would be ```isinstance(pk, models.Model)```. If we are not looking at a model object, getting the primary key is not useful. Furthermore, not everyone is using ```id``` as their primary key (hence the ```_get_pk_value``` method). If the primary key seems to be a model, just recursively call ```_get_pk_value```.

You can update your pull request with the proposed changes, but if it takes too long I might implement a solution myself so it can be shipped with the upcoming release.

johnrtipton:
I've updated it with your suggestion. This looks much cleaner now.  Thanks for the tip.

jjkester:
I am sorry to have dragged this PR for such a long time - of course it is good to merge. Thanks for your contribution!

PR #20, opened 2015-04-30
Add remote address to audit log

mrbaboon:
Still fleshing this out, but is there interest in this feature?


mrbaboon:
Added release v0.2.2 so I can pull it down from my fork for now.

jjkester:
Thank you for your contribution. Peronally I think this could be handy. I myself was thinking about adding a text field that could contain a serialized dictionary for user-defined metadata (including the remote address). However, I guess that the remote address can be important and therefore it should be possible to filter on it.

I myself have not made up my mind, I would be happy if there are any users of auditlog can comment on this idea.

@mrbaboon: Please let me know when you are happy with your code so I can review and merge it.

jjkester:
Since there has not been any activity on this PR I will close this PR and add the functionality myself. It will still be included in the upcoming 0.3 release.

PR #21, opened 2015-05-04
Additional field for metadata

annp89:
Addresses the issue where users want to store more information in log entry instaances. For example, model metadata or information of related fields (foreign key pks, atrributes, etc.)

This is implemented here by adding a JSON field <code>detailed_object_repr</code> to the model. On creation of log entries, if the model being logged has defined a function called <code>get_detailed_object_repr</code> all contents returned by that method gets stored in the <code>detailed_object_repr</code> field.


jjkester:
Thanks for your contribution. It would be nice if you could create a unit test for this (test the behavior of both a model that provides the ```get_detailed_object_repr``` method and one that doesn't).

annp89:
Sure thing. Will do!

jjkester:
I'm looking forward to it! I also looked at your reply to the original issue (#21) again and I think I prefer the _additional data_ name. That is indeed more flexible.

jjkester:
Hi,

I would like to include your PR into the next release, so it would be great if you can update your code soon. Thanks!

annp89:
I updated the naming and added tests. 
I also plan to open 2 more isolated pull requests (a)to ignore changes that get logged when None gets a default value of  '' and [] (b) Add support for comparing fields with JSON data.
Those are just so that we can talk about to see what your thoughts are. Thanks!

src/testapp/models.py
```diff
@@ -82,8 +82,33 @@ class SimpleExcludeModel(models.Model):
     history = AuditlogHistoryField()
 
 
+class AdditionDataIncludedModel(models.Model):
```
jjkester: I guess this is a typo. It does not break anything, but I think this class should be named ```AdditionalDataIncludedModel```.

src/auditlog/models.py
```diff
@@ -28,6 +30,10 @@ def log_create(self, instance, **kwargs):
             if isinstance(pk, (int, long)):
                 kwargs.setdefault('object_id', pk)
 
+            get_additional_data = getattr(instance, 'get_additional_data', None)
+            if callable(get_additional_data):
+                kwargs.setdefault('additional_data', instance.get_additional_data())
```
jjkester: I believe ```instance.get_additional_data()``` can simply be written as ```get_additional_data()``` as you defined that above.

src/auditlog/models.py
```diff
@@ -110,6 +116,7 @@ class Action:
     changes = models.TextField(blank=True, verbose_name=_("change message"))
     actor = models.ForeignKey(settings.AUTH_USER_MODEL, blank=True, null=True, on_delete=models.SET_NULL, related_name='+', verbose_name=_("actor"))
     timestamp = models.DateTimeField(auto_now_add=True, verbose_name=_("timestamp"))
+    additional_data = JSONField(blank=True, null=True)
```
jjkester: A verbose name (for example ```_("additional data")```) should be added to the field definition to make translation possible.

jjkester:
Thanks for your commits. I have commented on some of your code. Could you please review this? Then I'm happy to merge.

src/auditlog/models.py
annp89: You're right!

src/auditlog/models.py
annp89: Agreed!

src/testapp/models.py
annp89: Yes. It's a typo.

PR #25, opened 2015-06-01
Add support for comparing fields with JSON data.

annp89:
For example : 
<code>
a = {'c':'a','a':u'b'}
</code>
<code>
b = {'c':'a','a':'b'}
</code>
<code>
b == a // True
</code>
<code>
unicode(b) == unicode(a) // False
</code>

So, by checking if the type of the value being compared is **dict**, we can prevent **dict** values from getting converted to **unicode** for equality comparison.

### DIFF
--- a/src/auditlog/diff.py
+++ b/src/auditlog/diff.py
@@ -45,14 +45,18 @@ def model_instance_diff(old, new, **kwargs):
 
     for field in fields:
         try:
-            old_value = unicode(getattr(old, field.name, None))
+            old_value = getattr(old, field.name, None)
         except ObjectDoesNotExist:
             old_value = None
+        else:
+            old_value = unicode(old_value) if type(old_value) is not dict else old_value
 
         try:
-            new_value = unicode(getattr(new, field.name, None))
+            new_value = getattr(new, field.name, None)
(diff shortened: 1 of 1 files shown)


jjkester:
Good point. I think this might be an issue in more cases, especially with custom fields. This is obviously a flaw in my original code.

I think that always checking without string conversion is the way to go. Converting to a string should be done as late as possible. In that case the ```else``` part of the blocks can be removed and the string conversion should be done in line 62.

From the top of my head I cannot think of any cases where ```__eq__`` is not properly implemented for things that go into database fields. I also think that if people want to use Auditlog with any custom field they should themselves be responsible for correctly implementing ```__eq__``` as this is always a good thing to do. This can be put into the Auditlog documentation so this behavior is clear to users.

What do you think of this proposal?

jjkester:
It just occured to me that you are using an out-of-date base branch without Python 3 support. I recommend you to update your fork with the latest changes.

I decided to create a PR for my proposed change together with the default values fix as it all is about the same code.

jjkester:
I like to hear what you think of my PR, so please comment on it. I'll close this one in favor of #28.

PR #26, opened 2015-06-01
Prevent creating change logs when default values are set

annp89:
Safe to ignore changes from None to empty list/string

### DIFF
--- a/src/auditlog/diff.py
+++ b/src/auditlog/diff.py
@@ -54,6 +54,11 @@ def model_instance_diff(old, new, **kwargs):
         except ObjectDoesNotExist:
             new_value = None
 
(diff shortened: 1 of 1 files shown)


jjkester:
I think there is a more elegant way to do this. It is possible to query a model field for its default value, so it is possible to check against that instead of always checking against ```None```.

Django has a clear API for this now: https://docs.djangoproject.com/en/1.8/ref/models/meta/ . Only thing to check is that it works with 1.7 as well (see bottom of page for 1.7 API examples and their 1.8 counterparts).

I will close this PR and will add this functionality myself.

jjkester:
Thanks for pointing out the issue!

PR #34, opened 2015-08-20
Changed import statement and reference to GenericRelation in order

grendel513:
to remove the following Django Warning:
> site-packages/auditlog/models.py:6: RemovedInDjango19Warning:
> django.contrib.contenttypes.generic is deprecated and will be removed in Django 1.9. Its contents have been moved to the fields, forms, and admin submodules of django.contrib.contenttypes.
> from django.contrib.contenttypes import generic

### DIFF
--- a/src/auditlog/middleware.py
+++ b/src/auditlog/middleware.py
@@ -6,7 +6,7 @@ import time
 from django.conf import settings
 from django.db.models.signals import pre_save
 from django.utils.functional import curry
-from django.db.models.loading import get_model
+from django.apps import apps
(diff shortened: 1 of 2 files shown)


jjkester:
Ah, thanks for noticing. As far as I know Django 1.9 is still some time away, but it is always nice to get rid of warnings ;-)

I guess it is lucky that I dropped support for 1.6 and lower since this way of model loading is included since 1.7.

I have flagged your issue for the next minor release. I will take a closer look later when I have a bit more time to properly push a new release.

Thanks for contributing!

PR #38, opened 2015-10-02
Django 1.8 Compatibility

johnrtipton:
Fixed some deprecated calls.


jjkester:
Doesn't this tackle the same issues as #34? If so, why is your solution better?

johnrtipton:
Yes, it looks like we both patched this issue.  It does look like mine has more backwards compatibility checking, so you may want to look into that.  Otherwise, feel free to ignore this request.

jjkester:
Django 1.7 is currently the lowest supported version, so backwards compatibility is not really needed. I think I'll go with the other pull request simply because that one was earlier. Nevertheless, thanks for contributing.

johnrtipton:
Sure, no problem.

PR #41, opened 2015-10-18
Add db_index to object_pk

7wonders:
Was seeing a big hit on heroku because of object_pk not having an index. This seems to clean it up. The 0005 migration is happening automatically when I run makemigrations and doesnt really change anything.


jjkester:
After looking at the code I am not really surprised that this causes performance issues, the index is needed. Don't worry about the 0005 migrations, Django's migration stuff thinks a changed verbose name is worthy of a migration, so...

<code style nazi mode on> Could you put ```db_index=True``` before the verbose name? (Then it looks more like the other lines in the file.) <code style nazi mode off>

Looks good, thanks for noticing and immediately fixing the bug!

jjkester:
Thanks :-)

PR #43, opened 2015-11-11
Fix unregister call and add tests. Fixes #42

### LINKED ISSUE #42: AuditlogModelRegistry.unregister doesn't work
There are two issues:

1. iterating over the signals dict requires calling `.items()`
2. `signal.disconnect()` needs the same parameters as `signal.connect()`

kbussell:


jjkester:
Thanks for contributing. I think this will need to be unit tested at some point (metal note made).

jjkester:
A new version of django-auditlog, including your fix, has just been released. Thanks again for contributing!

PR #47, opened 2016-01-05
Change LogEntry.get_for_objects to distinguish between number and text keys.

justinlittman:
When using LogEntry.get_for_objects with a model that has an automatic (numeric) primary key, the following error occurs:
```
operator does not exist: text = integer LINE 1: ...ction_id" = 1) OR "auditlog_logentry"."object_pk" IN (SELECT... ^ HINT: No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

I suspect this is a Postgres-specific error.


jjkester:
Thanks for your contribution!

jjkester:
Just wanted to point out that a new version, including this fix, has just been released!

justinlittman:
Great timing and much appreciated.

PR #59, opened 2016-07-27
Change existing migration to fix issues for MySQL users

### LINKED ISSUE #54: Index column size too large on MySQL 5.7
`django.db.utils.OperationalError: (1709, 'Index column size too large. The maximum column size is 767 bytes.')`

Can be fixed by setting `max_length=255` in `0006_object_pk_index.py`.

jjkester:
Fixes #54.


jjkester:
Suffers from #60

PR #76, opened 2016-12-14
Added app name in audit log entry model

kapsonic:
app name can be useful to filter audit log entries by app

### DIFF
--- a/src/auditlog/models.py
+++ b/src/auditlog/models.py
@@ -37,6 +37,7 @@ class LogEntryManager(models.Manager):
             kwargs.setdefault('content_type', ContentType.objects.get_for_model(instance))
             kwargs.setdefault('object_pk', pk)
             kwargs.setdefault('object_repr', smart_text(instance))
+            kwargs.setdefault('app_name', instance._meta.app_label) # get the app name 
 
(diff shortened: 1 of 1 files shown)


jjkester:
Travis failed because there is no migration for this change (and therefore the database column does not exist).

While your use case is reasonable and realistic this field is not really needed. This is because a log entry is connected to a certain other model instance by using the [contenttypes framework provided by Django](https://docs.djangoproject.com/en/1.10/ref/contrib/contenttypes/). This means that there is a direct foreign key relation to an instance of `ContentType`, which already contains the app the model is in. So I really do not think this is needed anyway.

I do not want to discourage anyone from contributing of course, but I do not really see the point of having an app label in `LogEntry` as well.

To solve your problem, the following code will work perfectly:

```python
app_label = 'some_app'
LogEntry.objects.filter(content_type__app_label=app_label)
```

I am not sure whether Django allows setting an admin filter on the field `content_type__app_label`, but you might try.

kapsonic:
@jjkester It worked exactly as I expected.
Now that it is working, adding app_name column would be wrong.

Thanks for your suggestion.

PR #80, opened 2016-12-30
Updated Mixin to use get_full_name method

LexFrench:
Updated LogEntryAdminMixin to use models.User.get_full_name() method for django 1.10.

https://docs.djangoproject.com/en/1.10/ref/contrib/auth/#django.contrib.auth.models.User.get_full_name


chris-griffin:
@jjkester could you merge this? The admin portal doesn't work without this update. Thanks for all of your contributions on this app!

jjkester:
Thanks for the contribution. However, since then #85 opened which also contains this fix, and another, which puts me in a difficult position.

I have decided to merge #85 and not this PR because cherry picking the one fix from #85 means more work for me. Still, @LexFrench, thanks for your contribution, your fix is excellent.

PR #81, opened 2017-01-08
Integrate jsonfield_compat

### LINKED ISSUE #71: Adding Auditlog to Installed Apps causes PostreSQL JSON fields to be re-encoded on model save, without explicity registering the model
Following Django's documentation for PostgreSQL's JSON field: https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/fields/#querying-jsonfield

I get some weird behaviour after including AuditLog in my installed apps:

**#models.py**
```python
from django.contrib.postgres.fields import JSONField
from django.db import models


class Dog(models.Model):
    name = models.CharField(max_length=200)
    data = JSONField()

    def __str__(self):  # __unicode__ on Python 2
        return self.name
```

**#settings.py**
```python
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.postgresql_psycopg2',
        'NAME': 'audittest',
        'USER': 'audittest',
        'PASSWORD': 'audittest',
        'HOST': 'localhost',
        'PORT': '',
    }
}
```

**# pip freeze** 
```bash
(venv) jwe in ~/Documents/audittest λ pip freeze 
Django==1.10.1
django-auditlog==0.4.0
django-jsonfield==1.0.1
psycopg2==2.6.2
```

**./manage.py shell**
```python
(venv) jwe in ~/Documents/audittest λ ./manage.py shell
Python 3.4.2 (default, Oct  8 2014, 10:45:20) 
[GCC 4.9.1] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from jsontest.models import Dog
>>> Dog.objects.create(name='Meg', data={'breed': 'collie'})
<Dog: Meg>
>>> Dog.objects.first().data
'{"breed": "collie"}'
>>> Dog.objects.first().save()
>>> Dog.objects.first().data
'"{\\"breed\\": \\"collie\\"}"'
>>> Dog.objects.first().save()
>>> Dog.objects.first().data
'"\\"{\\\\\\"breed\\\\\\": \\\\\\"collie\\\\\\"}\\""'
>>> Dog.objects.first().save()
>>> Dog.objects.first().data
'"\\"\\\\\\"{\\\\\\\\\\\\\\"breed\\\\\\\\\\\\\\": \\\\\\\\\\\\\\"collie\\\\\\\\\\\\\\"}\\\\\\"\\""'
>>> 
```

**Things to note:**

-  I did not include `'auditlog.middleware.AuditlogMiddleware'`
- Issue occurs both before and after I run `./manage.py migrate`
- I did not register the model to log for changes with `auditlog.register(Dog)`

I'll try snooing around the code base this weekend to figure out what's going on, but I can't garuntee anything. Just thought I make this known incase any other people encouter the issue. 

Cheers, Jack

kbussell:
This commit integrates django-jsonfield-compat which aims to fix #71 .

After using this branch, add
```
USE_NATIVE_JSONFIELD = True
```
to your project's settings, and then run `python manage.py migrate`.

This will update auditlog's LogEntry model to be a native JSONField type. You can switch everything back by changing the above setting to `False`, just be sure to run migrations afterwards.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 Django>=1.8
 django-jsonfield>=1.0.0
+django-jsonfield-compat==0.4.4
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,7 @@ setup(
     install_requires=[
         'Django>=1.8',
         'django-jsonfield>=1.0.0',
+        'django-jsonfield-compat==0.4.4',
     ],
     zip_safe=False
 )
--- a/src/auditlog/apps.py
+++ b/src/auditlog/apps.py
@@ -6,3 +6,8 @@ from django.apps import AppConfig
 class AuditlogConfig(AppConfig):
     name = 'auditlog'
     verbose_name = "Audit log"
+
+    def ready(self):
+        import jsonfield_compat
+
+        jsonfield_compat.register_app(self)
--- a/src/auditlog/migrations/0004_logentry_detailed_object_repr.py
+++ b/src/auditlog/migrations/0004_logentry_detailed_object_repr.py
@@ -2,7 +2,8 @@
 from __future__ import unicode_literals
 
 from django.db import models, migrations
-import jsonfield.fields
+
+from jsonfield_compat import JSONField
 
 
 class Migration(migrations.Migration):
@@ -15,6 +16,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='logentry',
             name='additional_data',
-            field=jsonfield.fields.JSONField(null=True, blank=True),
+            field=JSONField(null=True, blank=True),
         ),
     ]
--- a/src/auditlog/migrations/0005_logentry_additional_data_verbose_name.py
+++ b/src/auditlog/migrations/0005_logentry_additional_data_verbose_name.py
@@ -2,7 +2,8 @@
 from __future__ import unicode_literals
 
 from django.db import migrations, models
-import jsonfield.fields
+
+from jsonfield_compat import JSONField
 
 
 class Migration(migrations.Migration):
@@ -15,6 +16,6 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='logentry',
             name='additional_data',
-            field=jsonfield.fields.JSONField(null=True, verbose_name='additional data', blank=True),
+            field=JSONField(null=True, verbose_name='additional data', blank=True),
         ),
     ]
--- a/src/auditlog/models.py
+++ b/src/auditlog/models.py
@@ -11,7 +11,7 @@ from django.utils.encoding import python_2_unicode_compatible, smart_text
 from django.utils.six import iteritems, integer_types
 from django.utils.translation import ugettext_lazy as _
 
-from jsonfield.fields import JSONField
+from jsonfield_compat import JSONField
 
 
 class LogEntryManager(models.Manager):


jjkester:
At this moment there is an issue with this solution on Python 3.

kbussell:
The package I uploaded to pypi originally was for python2 only. I've updated that now.

jjkester:
Great, it builds now as well :-)

I like to hear when you think it is mergable.

kbussell:
I finally got around to adding a wider range of tests to [django-jsonfield-compat](https://github.com/kbussell/django-jsonfield-compat), and am now comfortable with this being used a by a wider audience.

jjkester:
While I still am in favor of a compatibility layer I do want to see your particular solution before accepting these changes. You will hear from me soon.

jjkester:
I do not really like the fact that you do some SQL "fixes", can you elaborate a bit more on your implementation?

kbussell:
I chose to use direct DB manipulation so a simple Django setting could be used instead of having to create a migration for every model that used a JSONField.

The `register_app` call adds a post-migrate handler which may update the column type of all fields defined as `jsonfield_compat.JSONField`.

First, the column types are determined by [this query](https://github.com/kbussell/django-jsonfield-compat/blob/master/jsonfield_compat/convert.py#L18-L21).

We then check whether the database field type should be TEXT or JSONB depending on the `USE_NATIVE_JSONFIELD` setting.

If the DB field type needs to change, it runs the [update query](https://github.com/kbussell/django-jsonfield-compat/blob/master/jsonfield_compat/convert.py#L30-L35).

Let me know if you want me to go into more depth.

jjkester:
I do not agree with you on the migrations part, I think having migrations for these kind of things is a good thing. Hacking in SQL should - in my opinion - never be done except in one-time, custom migrations where it is unavoidable.

I do not feel comfortable imposing your solution on the users of this library since I do not feel comfortable with it myself, so I will not merge this pull request.

Still, the issue you try to fix here is somewhat important. Another solution is therefore appreciated, for example just a wrapper around the two classes which imports and returns the right class for the environment.

kbussell:
What user experience would you want for implementers who have already integrated django-auditlog and now want to switch to using the native JSONField implementation?

Assume their projects also include include django_jsonfield via other common libraries (e.g. django-activity-stream and django-geojson), and directly in their custom apps.

There can be no mixing of native with jsonfield, so all apps have to switch at once.

I'm motivated to build support to make this work, as our company is in the situation I mention above.

(We already have custom builds of the three libraries that make this change, but we don't want to maintain a fork forever.)

jjkester:
I think a settings switch is still permissable, but as stated above I am not a fan of executing SQL ALTER statements manually outside of migrations (it might very well be the only practical way though). I realize that creating a migration for an external library is not something you would want to do as well. Maintaining two versions is also not a very good solution. So you could say that I am conflicted.

ses4j:
FWIW, @kbussell PR working for us.  We had been using Postgres native JSONField across the project, and are just bringing in auditlog now, which caused all of our JSON fields to become strings.  Switching to this branch and adding `USE_NATIVE_JSONFIELD = True` was perfect for us.

However, my 2c: If I _had_ been using TEXT json fields, I'd be pretty furious if adding auditlog had the secret side effect of converting my entire database schema's JSON fields from TEXT to JSONB.  I think it would be nice if auditlog could detect that you're in this bad situation and tell you, and the USE_NATIVE_JSONFIELD switch is fine.  But effecting that migration has to be VERY explicit and very intentional on the part of the user.

ses4j:
@kbussell @jjkester I'm seeing an issue with the diffing of JSON columns, which may be caused by this change.  My JSON is coming up as "changed" but it's just that the keys changed orders. It seems `auditlog.diff.get_field_value` needs to be special-cased for JSON fields.

eriktelepovsky:
Hi. Is there any progress regarding this issue, please?

eriktelepovsky:
Hello people. Is there somebody who can look into this PR again, please? Native JSON fields of Postgres are pretty important nowadays and crucial for many libraries and projects.

audiolion:
Hey @eriktelepovsky , this PR came to a bit of stand-still because of how to proceed with the implementation, the core issue is that if we proceed with this implementation and you switch your database (e.g. sqlite -> postgres), the next time you run `python manage.py migrate` there would be sql statements run that alter your data to the new format, and could corrupt it or mess it up, and you would have no idea as a developer that it was going to happen.

@jjkester was suggesting that a settings switch like `AUDITLOG_USE_NATIVE_JSONFIELD` could be added instead so the developer has to make a conscious decision to run the SQL alterations and with the right docs should be able to make a smooth transition from a DB that doesn't support JSON to a DB that does (e.g. Postgres).

We would need to make or modify this PR to be based off the `settings` switch, and then do some testing where we take a Sqlite3 database using `django-jsonfield` and convert it to Postgres and see how it goes. We also need to ensure as @ses4j mentioned that the `diff` algorithm correctly handles both types in a sane manner, and then we need to write good docs about how you transition from a `django-jsonfield` db column to a native one. So I think due to the large scope, it just hasn't seen forward progress.

If people want to contribute to this issue I can help them along but I don't have the time to work on it right now and it is also not a problem I have in my own use-case for this library at work so I can't justify time at work to fixing it.

florencioq:
Hi @eriktelepovsky,

If you are in a hurry, use my fork:
git+https://github.com/florencioq/django-auditlog.git

It works together with django-jsonfield-compat as a workaround for this issue.

Just replace 'django-auditlog' by 'git+https://github.com/florencioq/django-auditlog.git' in your requirements.txt.

Take care,
José

eriktelepovsky:
Thank you @florencioq . Actually, @kbussell already did the same in his commit of his PR. I believe, the best possible solution would be a direct implementation of django-jsonfield-compat into django-jsonfield so any library using JSONField could gain from its purpose.

ses4j:
While this PR continues to stall, I made a v0.4.5 version of this if anyone wants it:

`pip install git+git://github.com/mathandpencil/django-auditlog.git@44a653d13f9077b7684727347517ea46c5ed7c95#egg=django-auditlog==0.4.5`

eriktelepovsky:
Please consider using models.JSONField since Django 3.1 that can be used on all supported database backends:

https://docs.djangoproject.com/en/3.1/releases/3.1/#jsonfield-for-all-supported-database-backends

aqeelat:
Since the minimum supported version is django 3.2, there is no need for this pr.
I vote we close it.
@hramezani

PR #84, opened 2017-01-11
Use get_full_name() instead of incorrect full_name attribute

### LINKED ISSUE #83: 'User' object has no attribute 'get_full'
The admin mixin uses an attribute that doesn't exist in the default django `User` model. It should use the get_full_name() method instead. Pull request coming.

chris-griffin:
Fixes #83


chris-griffin:
Just noticed this is a duplicate of #80

PR #86, opened 2017-01-11
Properly check timezone-aware DateTimeFields for changes

### LINKED ISSUE #82: DateTimeFields change if the timezone changes.
Hello!

I was trying to implement audits for some models we use, and some of those models use Django DateTimeFields. When I update any field in the model, the auditlog also creates a change entry for the DateTimeField, changing it from timezone UTC to UTC+1, but when Django accesses the model after that, it just gets converted back to UTC time. This makes it so that the DateTimeFields are always marked as changed, even if it is actually the same time. See the image for an example:

image

Kurocon:
Django converts the timezone of DateTimeFields to UTC when the field gets saved, when you update a model which includes a DateTimeField, and your server is running in another timezone. This results in the AuditLog thinking you changed the timestamp, while it actually is the same time, but in another timezone.

This PR adds a specific check in the model_instance_diff function for DateTimeField models (and any subclasses of it), which converts the old and new values to UTC before comparing them to see if they've actually changed. It also adds tests to see if the code works properly. This fixes #82.

The extra test_setting (USE_TZ) is added because timezone support is disabled if it is not specified, this setting enables it.

DateFields are not timezone-aware, so they do not need fixing.

### DIFF
--- a/src/auditlog/diff.py
+++ b/src/auditlog/diff.py
@@ -49,6 +49,34 @@ def get_fields_in_model(instance):
     return instance._meta.fields
 
 
+def get_field_value(obj, field):
+    """
+    Gets the value of a given model instance field.
+    :param obj: The model instance.
+    :type obj: Model
+    :param field: The field you want to find the value of.
+    :type field: Any
+    :return: The value of the field as a string.
+    :rtype: str
+    """
+    if isinstance(field, DateTimeField):
+        # DateTimeFields are timezone-aware, so we need to convert the field
+        # to its naive form before we can accuratly compare them for changes.
+        try:
+            value = field.to_python(getattr(obj, field.name, None))
+            if value is not None:
+                value = timezone.make_naive(value, timezone=timezone.utc)
+        except ObjectDoesNotExist:
+            value = field.default if field.default is not NOT_PROVIDED else None
+    else:
+        try:
+            value = smart_text(getattr(obj, field.name, None))
+        except ObjectDoesNotExist:
+            value = field.default if field.default is not NOT_PROVIDED else None
+
+    return value
+
+
 def model_instance_diff(old, new):
     """
     Calculates the differences between two model instances. One of the instances may be ``None`` (i.e., a newly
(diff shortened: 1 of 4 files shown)


jjkester:
Good spot. I usually run my servers on UTC (and I think many people with me) so I think that's why no-one ever noticed until now.

I have doubts about the changes to the diff function, there might be a cleaner way to do this (the provided solution is not very DRY so to say).

Nevertheless, kudo's for not only creating a fix but a regression test as well.

I'm not merging this in straight away, I am going to think about how to make the diff function a bit better. An idea I already have is to put the conversion logic (timezone stuff and setting the default value) in some other function. @Kurocon, I would like to hear your opinion on this (or if you have any other ideas).

Kurocon:
I agree with you that this solution is not very DRY. I didn't really have a better way of implementing it at the time, but I'll see if I can think of a nicer way.

Kurocon:
How is this implementation? I could have also moved the try/catch outside of the if statement, but I didn't because there might be a case where a field does not throw the `ObjectDoesNotExist` exception, or the handling of it needs to be different.

src/auditlog/diff.py
```diff
@@ -48,6 +49,34 @@ def get_fields_in_model(instance):
...
+    :type field: Any
+    :return: The value of the field as a string.
+    :rtype: str
+    """
+    if isinstance(field, DateTimeField):
+        # DateTimeFields are timezone-aware, so we need to convert the field
+        # to its naive form before we can accuratly compare them for changes.
+        try:
+            value = field.to_python(getattr(obj, field.name, None))
+            if value is not None:
```
jjkester: If you add the `isinstance(field, DateTimeField)` here, wouldn't that get rid of the duplication starting from line 71?

Kurocon: Yes, that would be possible, I already noted that in the comment above. I didn't choose to do that because in the future there might be other fields that do not throw an `ObjectDoesNotExist` exception, but that do need to be distinguished using an `isinstance`-check. If you think moving the try-catch is a better solution, that's fine too and I'll change it.

jjkester: Good point, did not catch that from your previous comment(s).

PR #87, opened 2017-01-18
 LogEntryAdminMixin throws Attribute error If a user misses full_name

fsniper:


jjkester:
Thanks for your contribution.

I have to comments to make regarding this solution, I would like to hear your thoughts about it.

1. I think, while a try/catch block works, an if-statement with `hasattribute`/`hasattr(obj, '__call__')` might be cleaner. Not sure because it will be two things to be checked for.

2. In your code you assume that there is an `email` field, however, while likely, a custom user still might not have this field. I would opt for just the string representation of the user object since that is much more likely to be implemented.

fsniper:
Hello @jjkester considering your review, I believe second approach would be a safer approach which would also give the developers more freedom. I will change the pull request accordingly asap.

PR #91, opened 2017-01-26
Use make_naive only if settings.USE_TZ is True.

fernandoe:
My Django app is using the setting USE_TZ=False and it is throwing the following error:

screen shot 2017-01-26 at 01 45 31

screen shot 2017-01-26 at 01 45 57

If I change the settings to True, it work like a charm (but I cant change it by now). We need to support applications that use this settings to False.

So, I added a verification to consider the timezone only if it is enable in the Django app.

### DIFF
--- a/src/auditlog/diff.py
+++ b/src/auditlog/diff.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+from django.conf import settings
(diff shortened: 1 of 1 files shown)


jjkester:
Great addition. I have never imagined to test something without that setting as I usually only see applications that use time zone aware datetimes.

PR #92, opened 2017-01-26
Fixed autologflush for django 1.10

goodmase:
NoArgsCommand was removed in Django 1.10. Switched to BaseCommand which defaults to no arguments. This fixes the management command autologflush in Django 1.10


jjkester:
Thanks for your contribution!

PR #94, opened 2017-01-29
Fixes #93 - Add 'changes_display_dict' property to 'LogEntry' model to display diff in a more human readable format

### LINKED ISSUE #93: Have changes_dict show get_<field_name>_display
If I have a model with some choices

```python

class MyModel(models.Model):
    STATUS_CHOICES = (
        ('a', 'ACTIVE'),
        ('d', 'DISABLED'),
    )
    status = models.CharField(choices=STATUS_CHOICES, max_length=1, default='d')
```

If I update my model with an auditlog attached to it and get the `changes_dict` for the log entry I will get

```
[ 'status': [u'd', u'a'] ]
```

Is there a way to get the models `get_status_display()` method for that to give

```
[ 'status': [u'DISABLED', u'ACTIVE'] ]
```

audiolion:
'changes_display_dict' currently handles fields with choices, long textfields and charfields, and datetime fields.

- Fields with choices display in the diff with their human readable text.
- Textfields and Charfields longer than 140 characters are truncated with an ellipsis appended.
- Datetime fields are rendered with the format 'Aug. 21, 2017, 02:25pm'
- Support for Postgres's `ArrayField` and `django-multiselectfield` for those not on Postgres

A new kwarg was added to `AuditlogModelRegistry` called `mapping_fields`. The kwarg allows the user to map the fields in the model to a more human readable or intuitive name. If a field isn't mapped it will default to the name as defined in the model. Partial mapping is supported, all fields do not need to be mapped to use the feature.

Tests added for 'changes_display_dict'
Tests added for 'mapping_fields' property

Updated Documentation

Resolves #93


jjkester:
Thanks for your contribution. Because the changes are bit more major than usual I'll mark this for version 0.5.0.

I do not have a lot of time to look through the changes now, so I'll come back to it later.

audiolion:
so if you want I can break this PR up into a few separate changes, one would be to add the mapping fields attribute. This allow the diff to take a database defined name `sku` and map it to something more readable `Product No.`. This is an opt in feature (to provide field mappings in your auditlog register function) and we can include a boolean argument to the `diff` function to turn off the field mappings if you want to get the actual database field names.

audiolion:
PR now only adds changes display dict functionality. If you are not in favor of this approach you can close the PR and I will just use my custom fork. It is my opinion that due to the need for using model meta attribute introspection to solve this problem, template tags are an inefficient solution for this problem.

audiolion:
With the updates you can uses `changes_display_dict` to get:
image

The current use of `changes_dict` does not map to human readable names and provides the actual database values like this:
image

The information from `changes_dict` is great for the programmer, what database field was changed, what database value was changed, however if you are displaying it to your end users or non-technical users who are using the admin site the human readable version helps a lot.

jjkester:
OK, you convinced me. I have no time today to actually look at everything but I will take a look soon.

src/auditlog/models.py
```diff
@@ -229,6 +233,62 @@ def changes_str(self, colon=': ', arrow=smart_text(' \u2192 '), separator='; '):
 
         return separator.join(substrings)
 
+    @property
+    def changes_display_dict(self):
```
jjkester: Can you inline-comment this? Than it is better understandable what it exactly does.

src/auditlog/models.py
```diff
@@ -257,6 +317,7 @@ def __init__(self, pk_indexable=True, **kwargs):
         kwargs['content_type_field'] = 'content_type'
         super(AuditlogHistoryField, self).__init__(**kwargs)
 
+
```
jjkester: Newline not needed here

src/auditlog_tests/models.py
```diff
@@ -120,11 +134,49 @@ class DateTimeFieldModel(models.Model):
...
+    GREEN = 'g'
+
+    STATUS_CHOICES = (
+        (RED, 'Red'),
+        (YELLOW, 'Yellow'),
+        (GREEN, 'Green'),
+    )
+
+    status = models.CharField(max_length=1, choices=STATUS_CHOICES)
+    multiselect = MultiSelectField(max_length=3, choices=STATUS_CHOICES, max_choices=3)
```
jjkester: Why are you explicitly testing with this third party library?

src/auditlog_tests/test_settings.py
```diff
@@ -19,7 +20,7 @@
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': 'auditlog_tests.db',
+        'NAME': 'auditlog_test.db',
```
jjkester: The 'app' is still called auditlog_tests, so why this change?

src/auditlog/models.py
audiolion: will do

src/auditlog/models.py
audiolion: yeah I think that was my editor adding it automatically on save ill remove

src/auditlog_tests/test_settings.py
audiolion: not needed change, ill revert

src/auditlog_tests/models.py
audiolion: I use this library and wanted auditlog to be compatible with it

jjkester: Right. If this is added to the tests I would also want the built-in PostgreSQL equivalent/replacement of it (https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/fields/#arrayfield) since it is likely that people store lists that way as well.

audiolion: yeah I am stuck on 1.8 so using it, I will add arrayfield support though, good idea!

jjkester:
The fixes do not seem to be pushed to this branch, so I cannot merge it. Can you do this so we can get the functionality in?

audiolion:
Hey @jjkester addressed all the review comments.

To support the request for Postgres's `ArrayField` I had to modify the test setup to run with multiple databases. In the process I found a bug where Auditlog saved LogEntry's to the default database always. I fixed this so that LogEntry's are saved to the database of the model they are attached to.

src/auditlog/diff.py
```diff
@@ -135,7 +135,7 @@ def model_instance_diff(old, new):
         new_value = get_field_value(new, field)
 
         if old_value != new_value:
-            diff[field.name] = (smart_text(old_value), smart_text(new_value))
+            diff[model_fields['mapping_fields'].get(field.name, field.name)] = (smart_text(old_value), smart_text(new_value))
```
jjkester: I would rather keep this dictionary with just field names and do some mapping elsewhere (for display only). In that case the default could be the verbose name of the field (if it still exists).

src/auditlog/models.py
```diff
@@ -232,6 +237,73 @@ def changes_str(self, colon=': ', arrow=smart_text(' \u2192 '), separator='; '):
...
+        for field_name, values in iteritems(self.changes_dict):
+            # try to get the field attribute on the model
+            try:
+                field = model._meta.get_field(field_name)
+            except FieldDoesNotExist:
+                # use the reverse field mapping to get the field
+                from auditlog.registry import auditlog
+                model_fields = auditlog.get_model_fields(model._meta.model)
+                reversed_field = model_fields['reverse_mapping_fields'].get(field_name)
+                field = model._meta.get_field(reversed_field)
```
jjkester: Again, a FieldDoesNotExist can be raised, for example when a model is changed.

src/auditlog/models.py
```diff
@@ -232,6 +237,73 @@ def changes_str(self, colon=': ', arrow=smart_text(' \u2192 '), separator='; '):
...
+                        values_display.append(choices_dict.get(value, 'None'))
+                    except:
+                        values_display.append(choices_dict.get(value, 'None'))
+            else:
+                for value in values:
+                    # handle case where field is a datetime type
+                    if "DateTime" in field.get_internal_type():
+                        try:
+                            value = parser.parse(value)
+                            value = value.strftime("%b %d, %Y %I:%M %p")
```
jjkester: I would rather see use of Django's l10n stuff for this, so that the format adapts to different settings.

src/auditlog/models.py
```diff
@@ -232,6 +237,73 @@ def changes_str(self, colon=': ', arrow=smart_text(' \u2192 '), separator='; '):
...
+                        except ValueError:
+                            pass
+                    # handle case where field is a date type
+                    elif "Date" in field.get_internal_type():
+                        try:
+                            value = parser.parse(value)
+                            value = value.strftime("%b %d, %Y")
```
jjkester: Again, l10n is preferred.

src/auditlog/models.py
```diff
@@ -232,6 +237,73 @@ def changes_str(self, colon=': ', arrow=smart_text(' \u2192 '), separator='; '):
...
+                        except ValueError:
+                            pass
+                    # handle case where field is a time type
+                    elif "Time" in field.get_internal_type():
+                        try:
+                            value = parser.parse(value)
+                            value = value.strftime("%I:%M %p")
```
jjkester: Again, l10n is preferred.

src/auditlog/diff.py
jjkester: I would not use the mapping here. I think keeping the database representation as close to internal representation as possible is a good idea, especially because later on we are using database values to retrieve model fields. Changing the mapping would completely destroy that capability for the changed fields. The actual field names are less likely to change.

The mapping could be applied when the dictionary for display is built up (see my comment there for details).

src/auditlog/models.py
```diff
@@ -232,6 +237,73 @@ def changes_str(self, colon=': ', arrow=smart_text(' \u2192 '), separator='; '):
...
+                        except ValueError:
+                            pass
+                    # check if length is longer than 140 and truncate with ellipsis
+                    if len(value) > 140:
+                        value = "{}...".format(value[:140])
+
+                    values_display.append(value)
+
+            changes_display_dict[field_name] = values_display
```
jjkester: Instead of the field name we might use the mapping here. Also, we might consider going with the verbose name if present. So in order of importance: mapping - verbose name - name (the last is only guaranteed to be present).

audiolion: Ok made this change, the order is mapping -> verbose name -> default verbose name. Django assigns a default verbose_name if none is provided so it is guaranteed to be always there.

>If this isn’t given, Django will use a munged version of the class name: CamelCase becomes camel case.
[ref](https://docs.djangoproject.com/en/1.11/ref/models/options/#verbose-name)

src/auditlog/diff.py
audiolion: Ah yes I agree, I reverted this change and now only change the name in the `changes_display_dict` property

src/auditlog/models.py
audiolion: Alright, modified to now rely on `USE_L10N` first and fallback on `settings.DATETIME_FORMAT`, `settings.DATE_FORMAT`, and `settings.TIME_FORMAT` which have defaults and can be set if the user desires.

src/auditlog/models.py
audiolion: Ok, with the update to `mapping_fields` not changing the `diff.py` output, I didn't need the `reverse_mapping_fields`. I have now removed that section, and in the case of `FieldDoesNotExist` fall back on the `field_name` auditlog has stored.

docs/source/usage.rst
```diff
@@ -95,6 +108,37 @@ your models is equally easy as any other field::
...
+            <td>{{ value.1|default:"None" }}</td>
+          </tr>
+        {% empty %}
+          <p>No history for this item has been logged yet.</p>
+        {% endfor %}
+        </tbody>
+      </table>
+    </div>
+
+If you want to display the changes in a more human readable format use the :py:class:`LogEntry`'s :py:attr:`changes_display_dict` instead. The :py:attr:`changes_display_dict` will translate ``choices`` fields into their human readable form, display timestamps in the form ``Jun. 31, 2017 12:55pm``, and truncate text greater than 140 characters to 140 characters with an ellipsis appended.
```
jjkester: maybe be explicit about the l10n support

PR #100, opened 2017-02-16
Don't crash if the linked object is not in the admin (fixes #99)

### LINKED ISSUE #99: Don't crash when the audited model doesn't have an admin page
Currently, auditlog tries to generate links to the referenced instance sections in the admin, but sometimes the instances aren't enabled in the admin. auditlog should probably catch that exception and not generate a link.

skorokithakis:


jjkester:
For now, I'll accept no test so we can get this out the door. @audiolion, thanks for your comments and @skorokithakis, thanks for your contribution!

audiolion:
No tests leads to problems 😢 

The PR imports `from django.urls.exceptions import NoReverseMatch` from [master](https://github.com/django/django/blob/master/django/urls/exceptions.py) which [doesn't exist on Django 1.8](https://github.com/django/django/tree/1.8.17/django/urls/exceptions.py)

and causes the server to not even startup

jjkester:
Autsj. The test do indeed not hit urls.py, so this really is an issue then...

On 1.8 the correct path seems to be `django.core.urlresolvers.NoReverseMatch` so this should be a quick fix then. Going to do it now.

jjkester:
Continuing conversation in #105

amiraliakbari:
If the user model is not enabled in admin, the `user_url` function will fail similarly. I think the same try-catch fix should be applied to that function too.

PR #105, opened 2017-02-16
Fix import error on NoReverseMatch

jjkester:


audiolion:
thank you! I was about to submit a PR but then I thought, I don't have a test for it so I would be a total hypocrite hahahaha

PR #111, opened 2017-03-21
Support models with UUID primary keys

### LINKED ISSUE #110: Manager methods do not support UUID primary keys
I am using UUID primary keys in an app and receiving the following error:

```
ProgrammingError: operator does not exist: character varying = uuid
LINE 1: ...g_logentry" WHERE "auditlog_logentry"."object_pk" IN (SELECT...
                                                             ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

This happens because it is trying to query `LogEntry.object_pk` (a char field) with my UUID primary key. This is happening in `LogEntryManager.get_for_object` and `LogEntryManager.get_for_objects`.

### Steps to reproduce:

```
class MyModel(models.Model):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)

LogEntry.objects.get_for_objects(MyModel.objects.all())
```

grantmcconnaughey:
This PR adds support for models with primary keys that use Django's UUIDField.

Fixes #110

### DIFF
--- a/src/auditlog/models.py
+++ b/src/auditlog/models.py
@@ -75,7 +75,7 @@ class LogEntryManager(models.Manager):
         if isinstance(pk, integer_types):
             return self.filter(content_type=content_type, object_id=pk)
         else:
-            return self.filter(content_type=content_type, object_pk=pk)
+            return self.filter(content_type=content_type, object_pk=smart_text(pk))
 
     def get_for_objects(self, queryset):
         """
@@ -94,6 +94,9 @@ class LogEntryManager(models.Manager):
 
         if isinstance(primary_keys[0], integer_types):
             return self.filter(content_type=content_type).filter(Q(object_id__in=primary_keys)).distinct()
+        elif isinstance(queryset.model._meta.pk, models.UUIDField):
+            primary_keys = [smart_text(pk) for pk in primary_keys]
+            return self.filter(content_type=content_type).filter(Q(object_pk__in=primary_keys)).distinct()
         else:
             return self.filter(content_type=content_type).filter(Q(object_pk__in=primary_keys)).distinct()
 
--- a/src/auditlog/models.py
+++ b/src/auditlog/models.py
@@ -90,7 +90,7 @@ class LogEntryManager(models.Manager):
             return self.none()
 
         content_type = ContentType.objects.get_for_model(queryset.model)
-        primary_keys = queryset.values_list(queryset.model._meta.pk.name, flat=True)
+        primary_keys = list(queryset.values_list(queryset.model._meta.pk.name, flat=True))
 
         if isinstance(primary_keys[0], integer_types):
             return self.filter(content_type=content_type).filter(Q(object_id__in=primary_keys)).distinct()
--- a/src/auditlog_tests/models.py
+++ b/src/auditlog_tests/models.py
@@ -1,3 +1,5 @@
+import uuid
+
 from django.db import models
 from auditlog.models import AuditlogHistoryField
 from auditlog.registry import auditlog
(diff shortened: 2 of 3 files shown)


src/auditlog/models.py
```diff
@@ -94,6 +94,9 @@ def get_for_objects(self, queryset):
 
         if isinstance(primary_keys[0], integer_types):
             return self.filter(content_type=content_type).filter(Q(object_id__in=primary_keys)).distinct()
+        elif isinstance(queryset.model._meta.pk, models.UUIDField):
```
jjkester: This check is significantly different from the one above, which is not great. What type of value does UUIDField produce?

grantmcconnaughey: It provides a `uuid.UUID` instance, which errors when compared with the string in the `object_pk` field. Therefore I have to convert the UUIDs to strings.

I would use the new [`Cast` function](https://docs.djangoproject.com/en/1.10/ref/models/database-functions/#cast), but it was added in Django 1.10, so the user of django-auditlog would have to use Django 1.10 for this to work.

jjkester: If that is the case you can also do the following check:

```python
isinstance(primary_keys[0], uuid.UUID)
```

This is more in line with the check before by explicitly checking the type of the primary key value.

Another solution might be to always call smart_text on non-integer primary keys, but that might be unnecessary in most cases. Not sure myself which is better, I am sure you - as user of UUID fields - can choose something sensible here.

grantmcconnaughey: @jjkester Yes, that is an option. However, I believe `primary_keys[0]` will fire a query, which I would like to avoid. We can get the type of the primary key from `_meta` without having to query a row and inspect its type that way.

jjkester: Well, we need the value later on, so the query would have to be done anyways. Django should cache this, so it should be no problem. However, if you think it means having an extra query you can test this locally and supply some evidence, I would be interested in that.

audiolion: agreed with @grantmcconnaughey, an extra query will be fired off, even using `values_list()` the queryset is still lazily evaluated. If you want to use `primary_keys` you would need to force evaluation, e.g.
```
primary_keys = list(queryset.values_list(queryset.model._meta.pk.name, flat=True))
```

then `primary_keys[0]` is a constant time call, but the issue here is that if for some reason the primary key isn't in the first slot we would get unexpected behavior, whereas `queryset.model._meta.pk` is explicitly getting the primary key defined for that model through the `Meta` options

Maybe the solution is updated the previous

jjkester: Right. Apparently I wrongly assumed it would evaluate the whole list.

I would propose just evaluating it - we need to do it at some point and that would make the original if statement a bit faster AND keep everything very readable. (And we don't have the issue mentioned by @audiolion about the explicit field name.)

In any case, since we are concerned with the type of the value it would be better to check the type of the value rather than the type of something else that provided the value.

grantmcconnaughey: Another improvement of checking `queryset.model._meta.pk` and not `primary_keys[0]` is that `primary_keys[0]` will error raise an `IndexError` if there are no records for the table where checking `queryset.model._meta.pk` will not.

audiolion: @grantmcconnaughey agreed that is the advantage of using the `Meta`, I understand @jjkester trying to keep the code consistent to improve readability, but the current implementation does cause issues in some cases. I don't know how you replace the first check to see if it is of type (int/long) though, you can do `isinstance(queryset.model._meta.pk, models.AutoField)` but you would have to handle the other cases as well.

audiolion:
I believe we can refactor the code to improve readability later which is the main concern for this PR, for now I am going to merge this in so that we can get the support out there. I think this conversation also exposed an optimization we can make by forcing the queryset evaluation of `primary_keys` so that we don't hit the db doing `primary_keys[0]`.

audiolion:
having issues with travis picking up the build, going

PR #117, opened 2017-06-08
Deprecate the use of field.rel except for django 1.8

### LINKED ISSUE #116: field.rel deprecation warning
We need to use `field.remote_field` instead of `field.rel` which is available in django=>1.9. @jjkester are you ok with dropping support for django 1.8? 

The relevant warning:
```py
if getattr(field, 'rel', None) is not None and field.rel.to == LogEntry:
/auditlog/diff.py:25: RemovedInDjango20Warning: Usage of field.rel has been deprecated. Use field.remote_field instead.
```

chris-griffin:
Fixes #116

### DIFF
--- a/src/auditlog/diff.py
+++ b/src/auditlog/diff.py
@@ -24,7 +24,11 @@ def track_field(field):
         return False
 
     # Do not track relations to LogEntry
-    if getattr(field, 'rel', None) is not None and field.rel.to == LogEntry:
+    if getattr(field, 'remote_field', None) is not None and field.remote_field.model == LogEntry:
+        return False
+
+    # 1.8 check
+    elif getattr(field, 'rel', None) is not None and field.rel.to == LogEntry:
         return False
 
     return True


src/auditlog/diff.py
```diff
@@ -24,7 +24,11 @@ def track_field(field):
         return False
 
     # Do not track relations to LogEntry
-    if getattr(field, 'rel', None) is not None and field.rel.to == LogEntry:
+    if getattr(field, 'remote_field', None) is not None and field.remote_field.model == LogEntry:
+        return False
+
+    # 1.8 check
+    elif VERSION == (1, 8) and getattr(field, 'rel', None) is not None and field.rel.to == LogEntry:
```
audiolion: you need to `import django` and do `if django.VERSION == (1, 8):`

chris-griffin: yup thanks ☕️

jjkester: I don't really like the explicit version check, wouldn't it be sufficient to just do the `getattr()` call?

chris-griffin:
@jjkester I removed the explicit Django check as requested. Should be good to go.

PR #123, opened 2017-06-21
add_data to allow injecting additional metadata

agfunder:
m2m and mptt tracking code depends on this small core change

### DIFF
--- a/src/auditlog/models.py
+++ b/src/auditlog/models.py
@@ -44,23 +44,21 @@ class LogEntryManager(models.Manager):
             # instance is inspected for get_additional_data() method, which 
             # is used to populate additional_data field
             # here we assume get_additional_data(), if present, returns a dict
-            #
+
             # additionally, dict "add_data" can be passed into log_create,
             # which is merged with results of get_additional_data()
-            #
             # merged dict is stored to additional_data field
             #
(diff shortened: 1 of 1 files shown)


agfunder:
If this is accepted, I can likely PR the m2m and mptt code as is.
If this is rejected or changed, the m2m and mptt code will need some changes.
So I decided to submit this first.

agfunder:
For example of how I would use add_data, see https://github.com/agfunder/django-auditlog/commit/95e78dd97268e00ff4a268d7f4ee51582e33e07f#diff-f5c12bf42e8cd5ef3a793400e11cd5dfR23

Basically, allows internal code to automatically add some useful metadata into additional_data field.

audiolion:
Ill be able to review this properly in a day or so when I can pull the changes and play around with it

agfunder:
Maybe tests only require change to test_model_without_additional_data

audiolion:
Reviewing this again, I believe the best thing to do is create a third party app that pairs with `django-auditlog` and provides support for `m2m` and `mptt` tracking. This library from the start did not support `m2m` relations.

If you decide to make a third party library extension to `auditlog` I will try and support where I can.

PR #126, opened 2017-06-29
Chore -- Update testing

audiolion:
PR brings an updated testing model utilizing tox to quickly run tests against different all compatible python versions for each [currently supported version of Django](https://www.djangoproject.com/download/#supported-versions). Also starts tracking code coverage for our tests by using the coverage package and integrating with the free for open source codecov website.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,1 @@
-Django>=1.8
 django-jsonfield>=1.0.0
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,7 @@
 .idea
 .project
 .pydevproject
+.coverage
+venv/
+.tox/
 local_settings.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,29 @@
+# Config file for automatic testing at travis-ci.org
 language: python
+
 python:
-  - "2.7"
-  - "3.4"
+  - "3.5"
+
 env:
-  - DJANGO_VERSION=1.8.*
-  - DJANGO_VERSION=1.9.*
-  - DJANGO_VERSION=1.10.*
-  - DJANGO_VERSION=1.11a1
-install:
-  - "pip install -r requirements.txt"
-  - "pip install Django==$DJANGO_VERSION"
-script: "python src/runtests.py"
-sudo: false
+  - TOX_ENV=py36-django-18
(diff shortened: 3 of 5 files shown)


audiolion:
failure for the project codecov is expected until this is merged into master

jjkester:
What exactly is the benefit of tox compared to specifying different Travis CI interpreters?

audiolion:
The reason is more to keep local testing and testing on the CI using the same tools. `tox` makes it very simple for a contributor to just `pip install -r requirements-test.txt` and then `tox` to locally test their changes against all versions of python and versions of Django we support (given that they have the given python versions installed). Including it as the way of testing in the CI keeps it consistent with the local testing environment.

Taken from [tox's homepage](https://tox.readthedocs.io/en/latest/)
>Tox is a generic virtualenv management and test command line tool you can use for:
> - acting as a frontend to Continuous Integration servers, greatly reducing boilerplate and merging CI and shell-based testing.

So for this case it is the merging of CI and shell-based testing.

jjkester:
Clear, looks like a nice tool then. I guess this would also speed up CI since you don't have to build a new environment every time.

requirements.txt
```diff
@@ -1,2 +1 @@
-Django>=1.8
```
jjkester: Why did you remove this dependency?

audiolion: So as far as I knew, the only reason to include it in the requirements there was for testing, but now with `tox` testing you don't want to install `Django>=1.8` you want each version of Django tested against (currently 1.8, 1.10, 1.11). `tox` will automatically download the necessary Django versions to test against, so we no longer need it here as it would just be installing an extra dependency that isn't used.

I do understand that to USE `auditlog` you need Django, but that is already covered by the [install_requires](https://github.com/jjkester/django-auditlog/blob/master/setup.py#L13) in the `setup.py` for people installing the package.

jjkester: True. However, not everyone seems to use setuptools, and `install_requires` only works with setuptools. Not really a problem to remove it I guess since it should be pretty clear that Django 1.8 or higher is required.

audiolion: I never thought about people not using setuptools, but agreed on that point, and agreed that using auditlog it seems pretty obvious you would want a supported version of Django to run it :)

PR #127, opened 2017-07-03
Fix the DecimalField and JSONField comparison bugfix from ses4j

legios89:
Hi,

First of all this project is awesome!
I just met with this Decimal comparison problem what ses4j https://github.com/jjkester/django-auditlog/issues/112 already mentioned and solved. I also had to add this lines in my local version to prevent this kind of bug, so I hope that you can merge this into the next release because I like to switch back to the official version, but without this I can't right now.


audiolion:
Hey thanks for the PR! Would you mind adding a test case for this that asserts the diff output is as expected?

audiolion:
Merging in #158

PR #130, opened 2017-07-04
Add management commands package to setup.py

### LINKED ISSUE #129: The auditlogflush command does not work
Hi, Unknown command: 'auditlogflush'
Type 'manage.py help' for usage.

audiolion:
We weren't actually packaging up the management command when we published to PyPI, fixes #129


jjkester:
Oops, that is quite bad.

PR #131, opened 2017-07-14
Fix the resource_url for UUID primary key

Snake575:
Fixed issue that was originally resolved on #90


Snake575:
Hi!, sorry for the delay... I wrote a test but couldn't make it work though.

This is the test I tried to implement on `UUIDPrimaryKeyModelModelTest`:

```python
def test_admin_resource_url(self):
    ma = LogEntryAdmin(LogEntry, admin.site)
    obj_log_entry = self.obj.history.first()
    self.assertNotEqual(self.ma.resource_url(obj_log_entry), obj_log_entry.object_repr)
```

The test requires a proper admin config on `auditlog_test`

audiolion:
ok ill take a look at it

audiolion:
this was fixed in #147

PR #136, opened 2017-08-17
New feature, specify the operation(create, update, delete) on each register model

mopitz199:
New feature. it allows you to specify the operations(create, update, delete) on each register.

Example: register(YourModel, exclude_fields=[], include_fields=[], operations=['update', 'create'])

This feature was made to reduce the records in the databse because not always we want to use all the signals in our model


audiolion:
Hi @mopitz199 thanks for the PR, could you add some test cases to demonstrate the feature?

PR #143, opened 2017-10-08
Adds hasattr check before threadlocal.auditlog access in middleware.set_actor

smokeyfish:
middleware.set_actor is inconsistent with other methods in checking for hasattr(threadlocal, 'auditlog')


audiolion:
hey @smokeyfish thanks for the PR! requested a review from jjkester as I am not too familiar with this part of the codebase. Everything looks good though, do you mind adding a test case for this and we can merge?

Issue #107, opened 2017-03-03
ValueError: astimezone() cannot be applied to a naive datetime


rtkseagray:
Hello, it appears that the fix applied to resolve https://github.com/jjkester/django-auditlog/issues/82 is triggering an error in our environment. I've included a partial stack trace below.

Django 1.8 is configured with the following timezone settings:
```
TIME_ZONE = 'UTC'
USE_TZ = True
```

Partial stack trace:

```
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 710, in save
    force_update=force_update, update_fields=update_fields)
  File "/usr/local/lib/python2.7/dist-packages/django/db/models/base.py", line 747, in save_base
    update_fields=update_fields, raw=raw, using=using)
  File "/usr/local/lib/python2.7/dist-packages/django/dispatch/dispatcher.py", line 201, in send
    response = receiver(signal=self, sender=sender, **named)
  File "/usr/local/lib/python2.7/dist-packages/auditlog/receivers.py", line 16, in log_create
    changes = model_instance_diff(None, instance)
  File "/usr/local/lib/python2.7/dist-packages/auditlog/diff.py", line 131, in model_instance_diff
    new_value = get_field_value(new, field)
  File "/usr/local/lib/python2.7/dist-packages/auditlog/diff.py", line 69, in get_field_value
    value = timezone.make_naive(value, timezone=timezone.utc)
  File "/usr/local/lib/python2.7/dist-packages/django/utils/timezone.py", line 376, in make_naive
    value = value.astimezone(timezone)
ValueError: astimezone() cannot be applied to a naive datetime
```

Our model definition:
```
class InventoryAsset(Model):
  created = DateTimeField()
```

In this case, it looks like our timestamp is already naive? I haven't yet checked out the code, but I suspect the quick fix is first checking to see if the value is already naive. I will keep this issue updated if I discover anything else.

rtkseagray:
In our case, we're setting the field via `datetime.utcnow()`, which generates a naive `datetime` object (I'm actually not sure we're doing this correctly now that I think about it), which is what's getting saved in the created attribute. I think it's a happy coincidence that our system only operates in UTC, and thus we don't notice any time skews.

jjkester:
[...]

PR #144, opened 2017-10-12
Bugfix #107

svenauhagen:
Hi,

this is a bugfix for #107 

Thanks
Sven


audiolion:
Can you write a test case for passing a naive datetime and asserting the outcome?

svenauhagen:
Yes, I will write a test later today and submit it too

src/auditlog_tests/test_settings.py
```diff
@@ -30,14 +30,6 @@
         'ENGINE': 'django.db.backends.sqlite3',
         'NAME': 'auditlog_tests.db',
     },
-    'postgres': {
```
audiolion: why did you remove the postgres database tests? We have some tests that test postgres functionality specifically, I can understand that when you ran locally you might not have had a postgres server / db setup, but we shouldn't be removing it

svenauhagen: Sorry, I did not see that I committed that file as well.
I removed it for local testing on my end.
Do you want me to submit a complete new PR?

svenauhagen:
Sorry for the wrong commit I rebased and it should be fine now

audiolion:
yeah that is totally fine, thanks for rebasing, you didn't have to, I squash the commits going in so it wouldn't matter too much, but better to do it this way all the same for clarity with git's history.

svenauhagen:
Hi,

did this fix get merged at all?

Best
Sven

audiolion:
sorry lost track of this, working on merging now

PR #145, opened 2017-10-15
Add missing python-dateutil from setup.py.

lamby:


lamby:
Looks like this was (silently) fixed in https://github.com/jjkester/django-auditlog/commit/f45fb55c8124a944be5d3b7b05fa1457baba1126#diff-2eeaed663bd0d25b7e608891384b7298 ?

audiolion:
yep, thanks for the PR though!

PR #147, opened 2017-11-11
Fixed resource_url to work with non-id models

delneg:
Didn't check that object_pk is always populated but i hope so :D
P.S. Produced null for custom user model with UUID pk for me


audiolion:
Could you elaborate on the error you ran into and what this PR here does to solve it?

AuditLog uses `object_id` to hold the primary key of the model it is related to, so long as that model has an integer primary key, if it does not, we use `object_pk` which holds a string version of the object's primary key, so a `slug` field or `uuid` field.

Perhaps what we should do here is check if `obj.object_id is None`, if so, `try obj.object_pk`?

delneg:
Here's logentry object:
```python
In [5]: LogEntry.objects.get(id=6).__dict__
Out[5]: 
{'_state': <django.db.models.base.ModelState at 0x109995f60>,
 'action': 1,
 'actor_id': None,
 'additional_data': None,
 'changes': '{"last_login": ["2017-11-11 14:39:35.149417", "2017-11-13 10:14:33.227361"]}',
 'content_type_id': 7,
 'id': 6,
 'object_id': None,
 'object_pk': '9fbb27b1-724f-459a-8cae-b313279de08a',
 'object_repr': 'delneg',
 'remote_addr': None,
 'timestamp': datetime.datetime(2017, 11, 13, 10, 14, 33, 233922, tzinfo=<UTC>)}
```
It produces the following link in the admin interface in column 'Resource'
`<td class="field-resource_url"><a href="/admin/users/user/None/change/">delneg</a></td>`
My custom user model is here:
```python
class User(AbstractUser):
    id = models.UUIDField(
        primary_key=True,
        default=uuid.uuid4,
        editable=False
    )
 
    # Have to set pk_indexable to false because of non-integer primary key (UUID)
    # see http://django-auditlog.readthedocs.io/en/latest/usage.html#object-history
    history = AuditlogHistoryField(pk_indexable=False)

    def __str__(self):
        return self.username

auditlog.register(User, exclude_fields=['last_login'])
```
Despite of pk_indexable=False, it still fails to work properly.

audiolion:
yeah `pk_indexable` will make the relation work, e.g. you can access the history, but it doesn't define how lookup is done. Lets switch the statement to

```
args = [obj.object_pk] if obj.object_id is None else [obj.object_id]
link = urlresolvers.reverse(viewname, args=args)
```

and that way we keep the default functionality and cover this case as well.

delneg:
Done

audiolion:
you originally modified line 34 relating to obj.object_id links, this commit is modifying the obj.actor link, I think you modified the wrong line with recent commit

delneg:
Fuuuuuuuu

audiolion:
Just need to get some tests added and can merge, I can write them but it will be a few days before I can get to it

delneg:
Ok, i'm too tired of writing tests so i'll just wait for them.
P.S. i've wrote a serializer & other stuff for Django Rest framework integration, including custom content_type insides (instead of model's id it returns it's content), please let me know if you would like me to commit it into this repo.

audiolion:
hm, id have to see what it looks like, might be easier to just make it into your own package you maintain and we link to it from the README?

np on the tests

delneg:
Basically, it's just a serializer and a viewset.

Serializer:
P.S. In order to get_content_type to work, the model has to define a get_serializer() classmethod.
```python

from auditlog.models import LogEntry
import json
from rest_framework import serializers
class LogSerializer(serializers.ModelSerializer):
    action = serializers.SerializerMethodField()
    content_type = serializers.SerializerMethodField()
    changes = serializers.SerializerMethodField()

    class Meta:
        model = LogEntry
        fields = '__all__'

    def get_action(self, obj):
        return obj.get_action_display()

    def get_content_type(self, obj):
        default = str(obj.content_type)
        model_class = obj.content_type.model_class()
        instance = model_class.objects.filter(pk=obj.object_pk).first()
        if instance:
            try:
                serializer = model_class.get_serializer()
            except AttributeError:
                return default
            if serializer:
                return dict(serializer(instance).data)
        return default

    def get_changes(self, obj):
        return json.loads(obj.changes)
```
Viewset
```python

from rest_framework import viewsets, mixins
from auditlog.models import LogEntry
class LoggingViewsSet(viewsets.GenericViewSet,
                      mixins.ListModelMixin,
                      mixins.RetrieveModelMixin):
    serializer_class = LogSerializer

    def get_queryset(self):
        """
        There's added filters that can take model name like "user"(the names you can see in a
        base_name attr in urls) and return logs with info about changes in this model
        """

        queryset = LogEntry.objects.order_by('timestamp').reverse()
        from_time = self.request.query_params.get('from', None)
        to_time = self.request.query_params.get('to', None)
        model_str = self.request.query_params.get('model', None)
        if model_str:
            try:
                model = ContentType.objects.get(model=int(model_str))
                queryset = LogEntry.objects.get_for_model(model.model_class())
            except Exception:
                pass # Handle this exception (model class not found or smth else)
        if from_time:
            queryset = queryset.filter(timestamp__gt=from_time)
        if to_time:
            queryset = queryset.filter(timestamp__lt=to_time)
        return queryset
```

Issue #153, opened 2017-12-05
Update django-auditlog for Compatibility with Django 2.0


rodmanning:
Just updated one of my projects to use Django 2.0 and a number of tests now failing due to the _django-auditlog_ codebase. Issues appear to be:
1. `on_delete` now a positional argument for `ForeignKey` fields, not a keyword argument
2. `django.core.urlresovlers` renamed `django.urls`
3. `is_authenticated()` method now cannot be called as a method (still callable as a property (i.e. `is_authenticated`)).

[Django 2.0 Release Notes](https://docs.djangoproject.com/en/2.0/releases/2.0/)

Have a pull request ready to go with these changes, just not sure how you'd like to handle the migrations (would you prefer to see them edited for compatibility or squashed — I'd go with squashed, myself).

PR #154, opened 2017-12-06
Changes Required for Compatibility with Django 2.0

rodmanning:
Changes as noted in #153 for compatibility with django v2.0. All tox tests I can run are passing (but `py27-django-111` is refusing to run for me due to a dependancy problem as `functools` apparently doesn't like me today ...)

Still have to add in the migrations — let me know if you want them sequential or squashed.

### DIFF
--- a/src/auditlog/compat.py
+++ b/src/auditlog/compat.py
@@ -6,10 +6,10 @@ def is_authenticated(user):
 
     Function provides compatibility following deprecation of method call to
     is_authenticated() in Django 2.0.
-
     """
-
-    if django.VERSION < (1, 10):
+    if not hasattr(user, 'is_authenticated'):
+        return False
+    if hasattr(user.is_authenticated, "__call__"):
         return user.is_authenticated()
     else:
         return user.is_authenticated
--- a/src/auditlog/compat.py
+++ b/src/auditlog/compat.py
@@ -5,11 +5,18 @@ def is_authenticated(user):
     Return whether or not a User is authenticated.
 
     Function provides compatibility following deprecation of method call to
-    is_authenticated() in Django 2.0.
+    `is_authenticated()` in Django 2.0.
+
+    This is *only* required to support Django < v1.10 (i.e. v1.9 and earlier),
+    as `is_authenticated` was introduced as a property in v1.10.
+
     """
     if not hasattr(user, 'is_authenticated'):
         return False
-    if hasattr(user.is_authenticated, "__call__"):
+    if callable(user.is_authenticated):
+        # Will be callable if django.version < 2.0, but is only necessary in
+        # v1.9 and earlier due to change introduced in v1.10 making
+        # `is_authenticated` a property instead of a callable.
         return user.is_authenticated()
     else:
         return user.is_authenticated
--- a/src/auditlog/compat.py
+++ b/src/auditlog/compat.py
@@ -1,18 +1,16 @@
 import django
 
 def is_authenticated(user):
-    """
-    Return whether or not a User is authenticated.
+    """Return whether or not a User is authenticated.
 
     Function provides compatibility following deprecation of method call to
     `is_authenticated()` in Django 2.0.
 
     This is *only* required to support Django < v1.10 (i.e. v1.9 and earlier),
-    as `is_authenticated` was introduced as a property in v1.10.
-
+    as `is_authenticated` was introduced as a property in v1.10.s
     """
     if not hasattr(user, 'is_authenticated'):
-        return False
+       return False
     if callable(user.is_authenticated):
         # Will be callable if django.version < 2.0, but is only necessary in
         # v1.9 and earlier due to change introduced in v1.10 making
--- a/src/auditlog/compat.py
+++ b/src/auditlog/compat.py
@@ -1,0 +1,15 @@
+import django
+
+def is_authenticated(user):
+    """
+    Return whether or not a User is authenticated.
+
+    Function provides compatibility following deprecation of method call to
+    is_authenticated() in Django 2.0.
+
+    """
+
+    if django.VERSION < (1, 10):
+        return user.is_authenticated()
+    else:
+        return user.is_authenticated
--- a/src/auditlog/middleware.py
+++ b/src/auditlog/middleware.py
@@ -41,7 +41,7 @@ class AuditlogMiddleware(MiddlewareMixin):
             threadlocal.auditlog['remote_addr'] = request.META.get('HTTP_X_FORWARDED_FOR').split(',')[0]
 
         # Connect signal for automatic logging
-        if hasattr(request, 'user') and hasattr(request.user, 'is_authenticated') and request.user.is_authenticated():
+        if hasattr(request, 'user') and hasattr(request.user, 'is_authenticated') and request.user.is_authenticated:
             set_actor = curry(self.set_actor, user=request.user, signal_duid=threadlocal.auditlog['signal_duid'])
             pre_save.connect(set_actor, sender=LogEntry, dispatch_uid=threadlocal.auditlog['signal_duid'], weak=False)
 
--- a/src/auditlog/middleware.py
+++ b/src/auditlog/middleware.py
@@ -42,7 +42,7 @@ class AuditlogMiddleware(MiddlewareMixin):
             threadlocal.auditlog['remote_addr'] = request.META.get('HTTP_X_FORWARDED_FOR').split(',')[0]
 
         # Connect signal for automatic logging
-        if hasattr(request, 'user') and hasattr(request.user, 'is_authenticated') and is_authenticated(request.user):
+        if hasattr(request, 'user') and is_authenticated(request.user):
             set_actor = curry(self.set_actor, user=request.user, signal_duid=threadlocal.auditlog['signal_duid'])
             pre_save.connect(set_actor, sender=LogEntry, dispatch_uid=threadlocal.auditlog['signal_duid'], weak=False)
 
--- a/src/auditlog/middleware.py
+++ b/src/auditlog/middleware.py
@@ -8,6 +8,7 @@ from django.db.models.signals import pre_save
 from django.utils.functional import curry
 from django.apps import apps
 from auditlog.models import LogEntry
+from auditlog.compat import is_authenticated
 
 # Use MiddlewareMixin when present (Django >= 1.10)
 try:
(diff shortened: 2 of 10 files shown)


audiolion:
hey, been on vacation, looks like everything here is done in a sane way and how I would do it. Gonna ping @jjkester for a review

src/auditlog/compat.py
```diff
@@ -0,0 +1,15 @@
+import django
+
+def is_authenticated(user):
+    """
+    Return whether or not a User is authenticated.
+
+    Function provides compatibility following deprecation of method call to
+    is_authenticated() in Django 2.0.
+
```
jjkester: Superfluous new line

src/auditlog/compat.py
```diff
@@ -0,0 +1,15 @@
...
+    """
+
+    if django.VERSION < (1, 10):
```
jjkester: This works perfectly, however, checking if `user.is_authenticated` is callable might work too.

No need to change this (obviously) unless you'd like to, kind of a personal preference I guess.

src/auditlog/middleware.py
```diff
@@ -41,7 +42,7 @@ def process_request(self, request):
             threadlocal.auditlog['remote_addr'] = request.META.get('HTTP_X_FORWARDED_FOR').split(',')[0]
 
         # Connect signal for automatic logging
-        if hasattr(request, 'user') and hasattr(request.user, 'is_authenticated') and request.user.is_authenticated():
+        if hasattr(request, 'user') and hasattr(request.user, 'is_authenticated') and is_authenticated(request.user):
```
jjkester: Maybe the attribute check can be combined with the compat function (and just return `False` if we don't know)? Just an idea, not sure if this would be practical.

src/auditlog/migrations/0001_initial.py
```diff
@@ -24,8 +24,8 @@ class Migration(migrations.Migration):
                 ('action', models.PositiveSmallIntegerField(verbose_name='action', choices=[(0, 'create'), (1, 'update'), (2, 'delete')])),
                 ('changes', models.TextField(verbose_name='change message', blank=True)),
                 ('timestamp', models.DateTimeField(auto_now_add=True, verbose_name='timestamp')),
-                ('actor', models.ForeignKey(related_name='+', on_delete=django.db.models.deletion.SET_NULL, verbose_name='actor', blank=True, to=settings.AUTH_USER_MODEL, null=True)),
-                ('content_type', models.ForeignKey(related_name='+', verbose_name='content type', to='contenttypes.ContentType')),
+                ('actor', models.ForeignKey(on_delete=django.db.models.deletion.SET_NULL, to=settings.AUTH_USER_MODEL, related_name='+', verbose_name='actor', blank=True, null=True)),
```
jjkester: Where does this change come from? It kind of makes no sense to me.

src/auditlog/migrations/0001_initial.py
```diff
@@ -24,8 +24,8 @@ class Migration(migrations.Migration):
...
+                ('content_type', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='contenttypes.ContentType', related_name='+', verbose_name='content type')),
```
jjkester: Any reason this migration is edited?

src/auditlog/compat.py
rodmanning: At the end of the block? Fixed/removed.

src/auditlog/compat.py
rodmanning: No problem to change it (tested change and all OK). 

Thought behind the original version was to make it clear why the code existed — that way when you drop support for <1.10 in the future it's more obvious what code can be removed from `compat.py`.

src/auditlog/middleware.py
rodmanning: As above, have changed as suggested and all tested OK.

Consideration was to avoid adding logic to `compat.py` so that it can eventually be removed when <1.10 isn't supported.

src/auditlog/migrations/0001_initial.py
rodmanning: 1. `content_type` requires `on_delete` kwargs in django 2.0 (see [Django 2.0 changelog](https://docs.djangoproject.com/en/2.0/releases/2.0/#features-removed-in-2-0))

```
File "/home/rod/src/git/django-auditlog.git/src/auditlog/migrations/0001_initial.py", line 30, in Migration
    ('content_type', models.ForeignKey(related_name='+', verbose_name='content type', to='contenttypes.ContentType')),
TypeError: __init__() missing 1 required positional argument: 'on_delete'
⋮
ERROR:   py34-django-20: commands failed
ERROR:   py35-django-20: commands failed
ERROR:   py36-django-20: commands failed
```
2. `actor` changes not intentional, so have reverted that line (think that might've been how it was auto generated by `manage.py makemigrations`, which I copied-pasted =)

Django 2.0 changelog suggests squashing migrations to manage this, and I'd support that approach.

src/auditlog/migrations/0001_initial.py
rodmanning: See below.

src/auditlog/compat.py
```diff
@@ -6,10 +6,10 @@ def is_authenticated(user):
 
     Function provides compatibility following deprecation of method call to
     is_authenticated() in Django 2.0.
-
     """
-
-    if django.VERSION < (1, 10):
+    if not hasattr(user, 'is_authenticated'):
+        return False
+    if hasattr(user.is_authenticated, "__call__"):
```
jjkester: It would be better to use `callable()` for this check: https://bugs.python.org/issue10518#msg122584

src/auditlog/migrations/0001_initial.py
jjkester: Thanks for the explanation. I'm also for squashing migrations, however, I'm not sure about the implications of squashing for existing installs. Should be a separate issue anyways I guess.

src/auditlog/compat.py
```diff
@@ -6,10 +6,10 @@ def is_authenticated(user):
 
     Function provides compatibility following deprecation of method call to
     is_authenticated() in Django 2.0.
```
jjkester: As per your comment, maybe add info here that this is only required for Django 1.10 and older.

src/auditlog/compat.py
audiolion: personally id prefer the `if django.VERSION < (1, 10)` code for the reason stated, it lets you know when you can drop it in the compat, whereas looking at the new code I have no idea why it was added or what versions of django it makes compatible without added comments.

src/auditlog/compat.py
rodmanning: Ok, no problem. Am away for a few days but will either see if I can edit it through github or do it when I’m back. 

(You’re sure you want to check callable() instead of the version?)

src/auditlog/migrations/0001_initial.py
```diff
@@ -25,7 +25,8 @@ class Migration(migrations.Migration):
                 ('changes', models.TextField(verbose_name='change message', blank=True)),
                 ('timestamp', models.DateTimeField(auto_now_add=True, verbose_name='timestamp')),
                 ('actor', models.ForeignKey(related_name='+', on_delete=django.db.models.deletion.SET_NULL, verbose_name='actor', blank=True, to=settings.AUTH_USER_MODEL, null=True)),
-                ('content_type', models.ForeignKey(related_name='+', verbose_name='content type', to='contenttypes.ContentType')),
+                ('content_type', models.ForeignKey(on_delete=django.db.models.deletion.SET_NULL, related_name='+', verbose_name='content type', to='contenttypes.ContentType')),
```
vshlapakov: @rodmanning There should be the same deletion mode (`CASCADE`, I guess) as in the model itself, otherwise it leads to an additional redundant migration.

rodmanning: Thanks! 👍 Didn't notice that — now fixed.

src/auditlog/compat.py
rodmanning: Ok. Added detail to both comment and in-line in the code as well. 

Note that this is only needed for 1.9 and earlier (django 1.10 accepts `is_authenticated` as a property ([1.10 release notes](https://docs.djangoproject.com/en/2.0/releases/1.10/#using-user-is-authenticated-and-user-is-anonymous-as-methods))).

src/auditlog/compat.py
rodmanning: Didn't hear back, so changed as suggested.

src/auditlog/compat.py
```diff
@@ -0,0 +1,22 @@
...
+def is_authenticated(user):
+    """
+    Return whether or not a User is authenticated.
+
+    Function provides compatibility following deprecation of method call to
+    `is_authenticated()` in Django 2.0.
+
+    This is *only* required to support Django < v1.10 (i.e. v1.9 and earlier),
+    as `is_authenticated` was introduced as a property in v1.10.
+
```
jjkester: Another one of those newlines

src/auditlog/migrations/0001_initial.py
```diff
@@ -25,7 +25,8 @@ class Migration(migrations.Migration):
                 ('changes', models.TextField(verbose_name='change message', blank=True)),
                 ('timestamp', models.DateTimeField(auto_now_add=True, verbose_name='timestamp')),
                 ('actor', models.ForeignKey(related_name='+', on_delete=django.db.models.deletion.SET_NULL, verbose_name='actor', blank=True, to=settings.AUTH_USER_MODEL, null=True)),
-                ('content_type', models.ForeignKey(related_name='+', verbose_name='content type', to='contenttypes.ContentType')),
+                ('content_type', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, related_name='+', verbose_name='content type', to='contenttypes.ContentType')),
+
```
jjkester: Superfluous newline

rodmanning: Removed.

src/auditlog/compat.py
rodmanning: Removed.

robguttman:
Thank you for this!  When can we expect a 0.4.5 with these changes?

jjkester:
I'll try to get it released this week.

robguttman:
Ping - any chance that a 0.4.5 can still happen this week?

jjkester:
Thanks for the ping. I had not forgotten about it. I'm just having a bit of a overwhelming week. I have just pushed 0.4.5 to PyPI, it should appear shortly on GitHub as well. Thanks for your patience!

robguttman:
Thanks, Jan-Jelle!  This allows us to upgrade to Django 2.0 fully now.

PR #155, opened 2018-01-02
(bugfix) changes_display_dict errors

audiolion:
Found some bugs with my previous pr to add `changes_display_dict` property.

This PR Fixes four bugs discovered with changes_display_dict, three that caused crashes and the other was datetimes being displayed in UTC instead of the servers set timezone.


audiolion:
Ok changed it to be a more terse syntax

PR #158, opened 2018-01-04
(bugfix) Add case for DecimalField so there aren't false alarm changes

### LINKED ISSUE #112: Comparison of Decimal types is flaky, sometimes returns false positive
I was seeing false positives on my decimal fields, where I "changed the value" from .17000 to .17.  Why does `get_field_value` stringify before comparing?  I added this in locally and it seemed happier:

```
    elif isinstance(field, (DecimalField, JSONField)):
        try:
        except ObjectDoesNotExist:
```

audiolion:
Closes #112

### DIFF
--- a/src/auditlog/diff.py
+++ b/src/auditlog/diff.py
@@ -2,7 +2,8 @@ from __future__ import unicode_literals
 
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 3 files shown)


audiolion:
no idea what is going on here. The tests work for me locally, but are failing on travis, I thought at first it was a database issue, but I tried with both sqlite and postgres and got the same result.

PR #159, opened 2018-01-12
wrap strings with mark_safe for django 2.0 support

vbabiy:


audiolion:
can you link me to the deprecation warning docs?

vbabiy:
>Support for the allow_tags attribute on ModelAdmin methods is removed.
https://docs.djangoproject.com/en/2.0/releases/2.0/

src/auditlog/mixins.py
```diff
@@ -1,6 +1,12 @@
 import json
 
 from django.conf import settings
+
+try:
+    from django.utils.safestring import mark_safe
+except ImportError:
```
audiolion: Why is this in a try/except? as far as I can tell `django.utils.safestring.mark_safe` is in Django 1.8, 1.9, 1.10, 1.11, and 2.0 with no plans for deprecation.

src/auditlog/mixins.py
```diff
@@ -27,9 +34,10 @@ def user_url(self, obj):
                 link = urlresolvers.reverse(viewname, args=[obj.actor.id])
             except NoReverseMatch:
                 return u'%s' % (obj.actor)
-            return u'<a href="%s">%s</a>' % (link, obj.actor)
+            return mark_safe(u'<a href="%s">%s</a>' % (link, obj.actor))
 
         return 'system'
+
     user_url.allow_tags = True
```
audiolion: remove this line as its not needed since we are adding `mark_safe`

src/auditlog/mixins.py
```diff
@@ -42,7 +50,8 @@ def resource_url(self, obj):
         except NoReverseMatch:
             return obj.object_repr
         else:
-            return u'<a href="%s">%s</a>' % (link, obj.object_repr)
+            return mark_safe(u'<a href="%s">%s</a>' % (link, obj.object_repr))
+
     resource_url.allow_tags = True
```
audiolion: remove this line as its not needed since we are adding `mark_safe`

src/auditlog/mixins.py
```diff
@@ -67,6 +77,7 @@ def msg(self, obj):
             value = [i, field] + (['***', '***'] if field == 'password' else changes[field])
             msg += '<tr><td>%s</td><td>%s</td><td>%s</td><td>%s</td></tr>' % tuple(value)
         msg += '</table>'
-        return msg
+        return mark_safe(msg)
+
     msg.allow_tags = True
```
audiolion: remove this line as its not needed since we are adding `mark_safe`

AceFire6:
Sorry if commenting here is inappropriate, however, I've recently solved this internally due to a recent move to Django 2.0

@vbabiy Instead of marking the returns as safe I believe a better solution would be to use [`format_html`](https://docs.djangoproject.com/en/2.0/ref/utils/#django.utils.html.format_html). The reason for this is that it conditionally escapes all arguments passed into `format_html` unless they are explicitly marked safe and returns a string that is already marked as safe. 

This is what I've done regarding this issue: [gist](https://gist.github.com/AceFire6/ad9d7943f9d0a2f825fc130f923372fe) (excuse the f-strings)

theskumar:
This PR should be closed now? @AceFire6

AceFire6:
@theskumar Yeah, #167 does the same things this does

PR #161, opened 2018-02-15
Fixed deprecation warnings showing up in Django 1.11

ksarge:
The backwards compatible code in diff.py and compat.py were still giving DeprecationWarnings in Django 1.11. I fixed the code so it won't call/reference the deprecated functions/properties. 

This PR fixes this issue here: https://github.com/jjkester/django-auditlog/issues/116

### DIFF
--- a/src/auditlog/diff.py
+++ b/src/auditlog/diff.py
@@ -24,12 +24,14 @@ def track_field(field):
         return False
 
     # Do not track relations to LogEntry
-    if getattr(field, 'remote_field', None) is not None and field.remote_field.model == LogEntry:
-        return False
+    if hasattr(field, 'remote_field'):
+        if getattr(field, 'remote_field', None) is not None and field.remote_field.model == LogEntry:
+            return False
 
     # 1.8 check
-    elif getattr(field, 'rel', None) is not None and field.rel.to == LogEntry:
-        return False
+    elif 'rel' in dir(field) and getattr(field, 'rel', None) is not None:
+        if field.rel.to == LogEntry:
+            return False
 
     return True
 
--- a/src/auditlog/diff.py
+++ b/src/auditlog/diff.py
@@ -25,13 +25,12 @@ def track_field(field):
 
     # Do not track relations to LogEntry
     if hasattr(field, 'remote_field'):
-        if getattr(field, 'remote_field', None) is not None and field.remote_field.model == LogEntry:
+        if field.remote_field is not None and field.remote_field.model == LogEntry:
             return False
 
     # 1.8 check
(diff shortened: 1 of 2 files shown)


src/auditlog/diff.py
```diff
@@ -24,8 +24,9 @@ def track_field(field):
         return False
 
     # Do not track relations to LogEntry
-    if getattr(field, 'remote_field', None) is not None and field.remote_field.model == LogEntry:
```
audiolion: why this change? the logic seems exactly the same, except `getattr()` was replace with `if hasattr` then the comparison.

Django switched their deprecation warnings to not be loud by default, [pulled from their release notes](https://docs.djangoproject.com/en/2.0/releases/1.11/#deprecating-warnings-are-no-longer-loud-by-default)

> This change allows third-party apps to support both Django 1.11 LTS and Django 1.8 LTS without having to add code to avoid deprecation warnings.

Our goal then isn't to add code to avoid deprecation warnings, it is not too long until April anyways when Django 1.8 support ends so I don't think we need to make the changes in this pr.

But let me know what your thoughts are on this and why you wanted to make the change, always happy to have people submitting prs and contributing!

audiolion:
On another note, Django switched their deprecation warnings to not be loud by default, [pulled from their release notes](https://docs.djangoproject.com/en/2.0/releases/1.11/#deprecating-warnings-are-no-longer-loud-by-default)

> This change allows third-party apps to support both Django 1.11 LTS and Django 1.8 LTS without having to add code to avoid deprecation warnings.

Our goal then isn't to add code to avoid deprecation warnings, it is not too long until April anyways when Django 1.8 support ends so I don't think we need to make the changes in this pr.

But let me know what your thoughts are on this and why you wanted to make the change, always happy to have people submitting prs and contributing!

src/auditlog/diff.py
ksarge: The logic is a little different, before it was possible for `field.remote_field` to exist, but not `field.remote_field.model == LogEntry`. This would cause field.rel to be referenced and sound off a deprecation warning.

ksarge:
Even though the documentation says that the warnings are silenced in 1.11, we're still getting deprecation warnings when we run our test suite. It's over 600 tests and the deprecation from auditlog generate us over 200 warnings. It makes it difficult to debug with all those warnings showing up in the output. 

Sure, we could wait until April to see support for 1.8 dropped, but I think this is a good intermediate solution that guarantees Django 1.8 code is not hit when running newer versions.

audiolion:
@ksarge sorry, I for some reason didn't get a notification about this. 1.8 is done now so its time to cleanup the 1.8 compat and make a new release for 1.11 and up that doesnt have warnings

PR #164, opened 2018-02-19
Configurable Logger Backend

mrbaboon:
Backends are now configurable in settings.

This enables the sending of logs directly to a configured python logger, avoiding database bloat when there are lots of modifications.  The default settings should leave everything existing untouched (defaults to ModelBackend), but we can now configure a logger backend to direct auditlogs to a configured logger.

I haven't included a log formatter, as that is up to the end user to decide how they want to format them (we'll be doing JSON, to be consumed by our ELK stack).   All of the audit metadata gets dumped into the logs `extra` for ease of extraction when formatting a log entry.

I've had to leverage a bit more data in thread storage as I didn't want to make another trip to the database to fetch the actor details when performing the audit logging.  Felt a little bad but let me know what you think.

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -5,3 +5,4 @@ tox>=1.7.0
 codecov>=2.0.0
 django-multiselectfield==0.1.8
 psycopg2
+mock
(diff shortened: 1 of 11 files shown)


audiolion:
Hey I will do a review tomorrow!

mrbaboon:
@audiolion going to close this and open a new PR of my our fork's master branch. - I've made some changes to setup.py as we're getting ready to move this into prod and are deploying off of our fork for the time being.

PR #165, opened 2018-02-22
Configurable Logger Backend

mrbaboon:
Backends are now configurable in settings.

This enables the sending of logs directly to a configured python logger, avoiding database bloat when there are lots of modifications. The default settings should leave everything existing untouched (defaults to ModelBackend), but we can now configure a logger backend to direct auditlogs to a configured logger.

I haven't included a log formatter, as that is up to the end user to decide how they want to format them (we'll be doing JSON, to be consumed by our ELK stack). All of the audit metadata gets dumped into the logs extra for ease of extraction when formatting a log entry.

I've had to leverage a bit more data in thread storage as I didn't want to make another trip to the database to fetch the actor details when performing the audit logging. Felt a little bad but let me know what you think.

Could definitely use some Readme/docs updates.  I can look into doing that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,11 @@
 from distutils.core import setup
 
+from setuptools import find_packages
+
 setup(
     name='django-auditlog',
     version='0.4.5',
-    packages=['auditlog', 'auditlog.migrations', 'auditlog.management', 'auditlog.management.commands'],
+    packages=find_packages(where='src', exclude=['*tests*', 'runtests']),
     package_dir={'': 'src'},
     url='https://github.com/jjkester/django-auditlog',
     license='MIT',
--- a/src/auditlog/backends/logger.py
+++ b/src/auditlog/backends/logger.py
@@ -1,0 +1,64 @@
+import logging
+
+from django.contrib.contenttypes.models import ContentType
+from django.utils.encoding import smart_text
+
+from auditlog import settings
+from auditlog.models import LogEntry
+from auditlog.backends.base import AuditBackend
+from auditlog.middleware import get_thread_data
+
+
+class LoggerBackend(AuditBackend):
+    """Python logger based backend for creation of audit log entries."""
+    message = '{action} on {content_type_name} {object_pk} by actor {actor_name}'
+    logger = 'audit.log'
+
+    def get_level(self):
+        """Returns the level """
+        return settings.AUDITLOG_LEVEL
+
+    def get_logger(self):
+        """Returns an the python logger for logging audit messages"""
+        return logging.getLogger(settings.AUDITLOG_LOGGER)
+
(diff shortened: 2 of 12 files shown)


setup.py
```diff
@@ -1,9 +1,11 @@
 from distutils.core import setup
 
+from setuptools import find_packages
+
 setup(
     name='django-auditlog',
     version='0.4.5',
-    packages=['auditlog', 'auditlog.migrations', 'auditlog.management', 'auditlog.management.commands'],
+    packages=find_packages(where='src', exclude=['*tests*', 'runtests']),
```
audiolion: I am ok switching to `find_packages` but I am concerned that the `exclude` is only excluding `src/auditlog_tests/tests.py` and not the entire `src/auditlog_tests`.

src/auditlog/backends/logger.py
```diff
@@ -0,0 +1,64 @@
...
+from auditlog import settings
+from auditlog.models import LogEntry
+from auditlog.backends.base import AuditBackend
+from auditlog.middleware import get_thread_data
+
+
+class LoggerBackend(AuditBackend):
+    """Python logger based backend for creation of audit log entries."""
+    message = '{action} on {content_type_name} {object_pk} by actor {actor_name}'
+    logger = 'audit.log'
```
audiolion: what does this do? I am not seeing references to `self.logger` and `audit.log` is set as the default in `settings.AUDITLOG_LOGGER`. Also, what is the `audit.log` default logger name related to? Any reason it needs to be a descendent of `audit`?

mrbaboon:
Thanks @audiolion just saw this review.

I'll plug away at these when I get time at work over the next few days.

setup.py
mrbaboon: I can tweak it.

src/auditlog/backends/logger.py
mrbaboon: I think it was legacy - was using during development but then I moved to a package setting.

mrbaboon:
PS. great review!

dsever:
Hi, 

Is this in plan to me merged?

mrbaboon:
@dsever we can close it or somebody else will have to pick it up.  Unfortunately I haven't been able to keep it up to date.

dsever:
Now I'm very sad :( I would like to have this feature

mrbaboon:
@dsever feel free to update the PR and take it over :D  We aren't using this package for audit logging anymore so I can't validate the time to get it back on track to merge.

cansin:
Ah, this is exactly what I was researching for as well. i.e. bypassing DB calls in favor of directly hitting a logger. @hramezani do you know if this is in works somewhere?

PR #167, opened 2018-03-01
Mark LogEntryAdminMixin methods output as safe where required

AceFire6:
As per my comment (https://github.com/jjkester/django-auditlog/pull/159#issuecomment-361938038). I believe using [`format_html`](https://docs.djangoproject.com/en/2.0/ref/utils/#django.utils.html.format_html) to be a better solution to that proposed in #159.

* Remove `allow_tags` as they've been deprecated as of Django 2.0
* Use `format_html` to provide conditional escaping and `mark_safe` functionality

### DIFF
--- a/src/auditlog/mixins.py
+++ b/src/auditlog/mixins.py
@@ -65,10 +65,7 @@ class LogEntryAdminMixin(object):
         msg = '<table><tr><th>#</th><th>Field</th><th>From</th><th>To</th></tr>'
         for i, field in enumerate(sorted(changes), 1):
             value = [i, field] + (['***', '***'] if field == 'password' else changes[field])
-            msg += format_html(
-                '<tr><td>{}</td><td>{}</td><td>{}</td><td>{}</td></tr>',
-                *value,
-            )
+            msg += format_html('<tr><td>{}</td><td>{}</td><td>{}</td><td>{}</td></tr>', *value)
 
         msg += '</table>'
         return mark_safe(msg)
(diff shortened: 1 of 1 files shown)


jjkester:
@robguttman Releases are a bit ad-hoc currently, I might investigate whether it is feasible to release automatically by creating a Travis job for that to take the manual interaction out of the equation.

As for now, I'm all for a new release, but I'm traveling for work at the moment, I'll see what I can do, but don't expect a release in a couple of days. Thanks for your patience.

robguttman:
@jjkester, understood - we appreciate your efforts on this great lib!

jjkester:
@robguttman Thanks for the heads-up. I created a PR (#183) to automatically deploy to Travis. I could not merge it right away due to failing tests. These need to be addressed, and I kind of hoped after a try myself that someone else would pick it up since my time is very limited right now.

Since I'm (still) traveling (again), I don't have the credentials to manually release a version available right now. If you want to look at the failing tests, feel free, so we can get #183 merged and the releases going. Otherwise you would have to wait since I won't be able to release manually for a couple of weeks.

robguttman:
@jjkester, the Travis tests look like they all passed.  What seems to have "failed" are the code coverage targets - but perhaps I'm reading the results wrong?  When I try to look at the codecov report I just see this message:
```
Missing base report.
Unable to compare commits because the base of the compare did not upload a coverage report.
```

What next to release 0.4.6?

texnic:
What is the current status? Any changes since last autumn?

PR #176, opened 2018-05-30
Indicate Python 3 support in metadata

Flimm:
This is useful so that websites like djangopackages.org can detect that this package supports Python 3.

https://djangopackages.org/packages/p/django-auditlog/


Flimm:
Thank you!

PR #181, opened 2018-08-02
support for ArrayField, fix of TypeError: unhashable type: 'list'

eriktelepovsky:


audiolion:
Hi @eriktelepovsky thanks for the PR, could you include a test for this change? Thanks 👍

auditlog/migrations/0004_logentry_detailed_object_repr.py
```diff
@@ -1,4 +1,10 @@
-from django.db import migrations, models
+try:
+    # Django 3.1
```
alieh-rymasheuski, line 2: I don't think we officially support Django pre-3.2.

auditlog/models.py
```diff
@@ -452,8 +460,13 @@ def changes_display_dict(self):
                             )
                         else:
                             values_display.append(choices_dict.get(value, "None"))
-                    except Exception:
+                    except ValueError:
                         values_display.append(choices_dict.get(value, "None"))
+                    except Exception:
+                        if isinstance(value, list):
```
alieh-rymasheuski, line 466: Can we move this `isinstance` check to line 455 (where we test `if type(value) is list` rather than `if isinstance(value, list)`?

auditlog/settings.py
```diff
@@ -0,0 +1,3 @@
+from django.conf import settings
+
+ENABLE_ELLIPSIS = getattr(settings, "AUDITLOG_ENABLE_ELLIPSIS", True)
```
alieh-rymasheuski, line 3: Should this change be a part of this pull request?

alieh-rymasheuski (changes requested):

alieh-rymasheuski:
I wonder why this pull request appeared in my Github inbox. Apparently, it was because of an unrelated push to `eriktelepovsky:master`.

aqeelat:
@hramezani I really don't understand the changes in this PR. It solves nothing. I think this PR should be closed.

hramezani:
Going to close it because it is too old

PR #190, opened 2018-09-24
Fix/better handling of deprecated functionality

seawolf42:
Currently the code is compatible with newer versions of Django with support back to 1.8; however, unit tests that trigger some functionality in django-auditlog cause Django to fire deprecation warnings, which can impact build systems that treat warnings as errors. This PR changes the approach to diff determination and the compat version of `is_authenticated()` to prevent these warnings from happening.

### DIFF
--- a/src/auditlog/compat.py
+++ b/src/auditlog/compat.py
@@ -1,5 +1,10 @@
 import django
-from django.utils.deprecation import CallableFalse, CallableTrue
+
+try:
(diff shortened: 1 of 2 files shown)


seawolf42:
Looks like I have some errors in there around different versions of Django, I'll fix those and amend the PR accordingly shortly, but please let me know if the PR is otherwise encouraged or discouraged if you get a chance.

PR #194, opened 2018-11-21
Use django postgres jsonfield

AceFire6:
Edit old migrations so people using this package for the first time won't have to have the old jsonfield dependency.


AceFire6:
Sorry, this was made in the wrong place

PR #202, opened 2019-03-15
Added support for tracking FK ID updates.

barqshasbite:
In Django, you can update the raw ID of a ForeignKey relation
using the '*_id' field without assigning it a new model. This
needs to be logged by the audit framework. Currently it does
not detect these changes.


aqeelat:
@hramezani do you think this and #280 are related?

barqshasbite:
This change is no longer valid with the latest code base. A new pull request has been created to address the issue in latest master.

PR #206, opened 2019-05-03
Fix tests for multiple database support.

varuna-sd:


kbussell:
Thank you for this patch, @varuna-sd. The commit was included in #212 and merged in

PR #217, opened 2019-11-12
Fix TOXENV variable name in deploy step

kbussell:
This should hopefully get the release to pypi working!


jjkester:
Wow, I took a look yesterday but did completely overlook the underscore...

PR #222, opened 2019-11-26
Supply PyPI credentials via environment variables

jjkester:


jjkester:
Environment variables have been set using a project specific PyPI token.

PR #237, opened 2020-02-14
[TASK] Support newer versions of python-dateutil

krukas:
python-dateutil only drops support for python 2.6 and 3.2 in python-dateutil version 2.7.0. This makes python-dateutil 2.7.0 and 2.8.0 still compatible with auditlog.

In one of mine envrionments python-dateutil 2.8.0 is installed and django-auditlog works with this, only pip will complain every time that versions are not compatible.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ setup(
     description='Audit log app for Django',
     install_requires=[
         'django-jsonfield>=1.0.0',
-        'python-dateutil==2.6.0'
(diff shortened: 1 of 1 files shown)


Linkid (approved):

Linkid:
Arg, already done in https://github.com/jazzband/django-auditlog/commit/33fa2490714954b189e24963e4586f86ce4f99c2. Thank you anyway.

Issue #245, opened 2020-04-27
Users can't remove their own account because of actor field


PabloCastellano:
Hello. I'm using the middleware for automatic registering changes in my models and I'm having troubles implementing a feature so that users can remove their own account.

I've traced it down and the issue happens when the middleware tries to set the actor that took the action, which is the user that is being removed (`request.user.delete()`):

https://github.com/jjkester/django-auditlog/blob/c3b06045e04655878dcb6125ae4baf9e51b7bb7c/src/auditlog/middleware.py#L81-L82

This results in the following exception:

```
django.db.utils.IntegrityError: insert or update on table "auditlog_logentry" violates foreign key constraint "auditlog_logentry_actor_id_959271d2_fk_auth_user_id"
DETAIL:  Key (actor_id)=(3) is not present in table "auth_user".
```

What is the best way to disable the pre_save signal for this view or how can I just skip setting the actor for this request?

Thanks!

mcanu:
I came across the same issue here, I used the `disable_auditlog` context manager:
```python
from auditlog.context import disable_auditlog

with disable_auditlog():
    request.user.delete()
```

blopker:
In case anyone else runs into this, I've extended @mcanu's solution with also logging the deletion manually:

```
        # make a log entry for the deletion
        LogEntry.objects.log_create(user, force_log=True, action=LogEntry.Action.DELETE).save()
        with disable_auditlog():
            user.delete()
```

PR #246, opened 2020-04-27
docs: add troubleshooting section

PabloCastellano:
Related to #245


jjkester:
I don't think the proposed solution in the docs page is something that people should be doing.

Not automatically logging changes to the `User` would be preferred. Improvements are coming, see #262

PR #249, opened 2020-05-10
Added ability to define many to many relationships and register them

Kingjmk:
Added ability to define many to many relationships and register them in auditlog with new signal and log creator as the default implementation does not connect post_add, post_remove and post_clear actions from m2m_changed django signal.

### DIFF
--- a/src/auditlog/migrations/0008_action_choices.py
+++ b/src/auditlog/migrations/0008_action_choices.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('auditlog', '0007_object_pk_type'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='logentry',
+            name='action',
+            field=models.PositiveSmallIntegerField(verbose_name='action', choices=[(0, 'create'), (1, 'update'), (2, 'delete'), (3, 'added')]),
+        ),
+    ]
--- a/src/auditlog/models.py
+++ b/src/auditlog/models.py
@@ -53,6 +53,7 @@ class LogEntryManager(models.Manager):
             # Delete log entries with the same pk as a newly created model. This should only be necessary when an pk is
             # used twice.
             if kwargs.get('action', None) is LogEntry.Action.CREATE:
+                # Don't delete many to many records related to this object
                 if kwargs.get('object_id', None) is not None and self.filter(content_type=kwargs.get('content_type'), object_id=kwargs.get('object_id')).exists():
                     self.filter(content_type=kwargs.get('content_type'), object_id=kwargs.get('object_id')).delete()
                 else:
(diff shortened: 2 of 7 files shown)


rawad1989:
@Kingjmk Nice change, I have stuck at this and finally I found a good solution to do it, this works with me and I'm looking forward to have this feature in the next release.

@kbussell  @vfigueiro  @skorokithakis  Can you please check this PR and validate it since it was a huge issue to log changes in many to many relations changes

ruteckimikolaj:
Very nice feature. I am looking forward to see it in the next release. I have implemented it in our project ( test env ) and we hope to use it in production.

we haven't find any bugs in code, yet.

Kingjmk:
> Very nice feature. I am looking forward to see it in the next release. I have implemented it in our project ( test env ) and we hope to use it in production.
> 
> we haven't find any bugs in code, yet.

Can i ask about your use case for this feature, as i used it for logging changes to on user permissions and groups maybe we can improve it to better suit more uses.

ruteckimikolaj:
> 
> 
> > Very nice feature. I am looking forward to see it in the next release. I have implemented it in our project ( test env ) and we hope to use it in production.
> > we haven't find any bugs in code, yet.
> 
> Thanks for your reply!
> 
> Can i ask about your use case for this feature, as i used it for logging changes to on user permissions and groups maybe we can improve it to better suit more uses.

We are using it for the same purpose. Logging user permission changes and group changes in Wagtail cms project.

src/auditlog/migrations/0008_action_choices.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('auditlog', '0007_object_pk_type'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='logentry',
+            name='action',
+            field=models.PositiveSmallIntegerField(verbose_name='action', choices=[(0, 'create'), (1, 'update'), (2, 'delete'), (3, 'added')]),
```
weilu, line 17: Probably should say 'add' instead of 'added' just to be consistent with the rest of the verb tense here

src/auditlog/models.py
```diff
@@ -62,6 +63,60 @@ def log_create(self, instance, **kwargs):
...
+            if isinstance(pk, integer_types):
+                kwargs.setdefault('object_id', pk)
+
+            get_additional_data = getattr(instance, 'get_additional_data', None)
+            if callable(get_additional_data):
+                kwargs.setdefault('additional_data', get_additional_data())
+
+            kwargs.setdefault('action', action)
+
+            instance_verbose_name = str(smart_text(instance._meta.verbose_name))
```
alieh-rymasheuski, line 96: This is the wrong object to use. Let's say there is the following model:
```python
class Tag(Model):
    name = CharField(...)

class User(Model):
     random_tags = ManyToManyField(Tag)
```
When the tags of a user change, `instance` here is the instance of `User`, so `instance_verbose_name` is `"user"`, which results in the following log entry:
```json
{"changes": {"user": ["Added", ["1 - tag1", "2 - tag2"]]}}
```
This is clearly wrong, as one would expect to see `"random_tags"` is the key in the changes dict here.

src/auditlog/registry.py
```diff
@@ -92,7 +98,28 @@ def _connect_signals(self, model):
...
+                    if 'post_save' in m2m_signals:
+                        post_save_receiver = self._signals[post_save]
+                        post_save.connect(post_save_receiver, sender=m2m_model, dispatch_uid=self._dispatch_uid(m2m_model, model))
+
+                    if 'pre_save' in m2m_signals:
+                        pre_save_receiver = self._signals[pre_save]
+                        pre_save.connect(pre_save_receiver, sender=m2m_model, dispatch_uid=self._dispatch_uid(m2m_model, model))
+
+                    setattr(m2m_model, '_map_signals', m2m_signals)
+                    signal.connect(receiver, sender=m2m_model, dispatch_uid=self._dispatch_uid(signal, m2m_model))
```
alieh-rymasheuski, line 120: This `dispatch_uid` is unique per dependency, not per a pair of dependent + dependency. Consider this code:
```python
class Tag(Model):
    name = CharField(...)

class User(Model):
    random_tags = ManyToManyField(Tag)

class Cat(Model):
    color_tags = ManyToManyField(Tag)

auditlog.register(User, m2m_fields={"random_tags": []})
auditlog.register(Cat, m2m_fields={"color_tags": []})
```
In this case we won't track the changes to `Cat.color_tags`, because its handler will have the same `dispatch_uid` as the one for `User.random_tags` (please correct me if I misunderstand the logic of `dispatch_uid`).

A better way would be to generate `dispatch_uid` from a pair of `User, Tag` and `Cat, Tag` respectively.

src/auditlog/registry.py
```diff
@@ -92,7 +98,28 @@ def _connect_signals(self, model):
         """
         for signal in self._signals:
             receiver = self._signals[signal]
-            signal.connect(receiver, sender=model, dispatch_uid=self._dispatch_uid(signal, model))
+
+            if signal is m2m_changed:
```
alieh-rymasheuski, line 102: This special logic doesn't have its counterpart in `_disconnect_signals`. Disconnect just won't work for m2m signals.

src/auditlog/receivers.py
```diff
@@ -61,3 +61,36 @@ def log_delete(sender, instance, **kwargs):
...
+                LogEntry.objects.log_m2m_changes(
+                    changed_queryset,
+                    kwargs['instance'],
+                    LogEntry.Action.ADDED,
+                )
+            elif action == 'post_remove':
+                LogEntry.objects.log_m2m_changes(
+                    changed_queryset,
+                    kwargs['instance'],
+                    LogEntry.Action.DELETE,
```
alieh-rymasheuski, line 89: It doesn't feel right to reuse `DELETE` action for this case. A `DELETE` log entry now shows the final operation on an object, while after this PR a `DELETE` log entry will either represent the deletion of the object itself or the deletion of any of the values in one of its fields.

A better choice would be to either add a new action (e.g. `M2M_DELETE`) or use `UPDATE` action for both addition and removal of items from an m2m field.

alieh-rymasheuski (changes requested):
This pull request has a few cosmetic issues (e.g. a bad choice of `action` or a wrong key in the changes dict) as well as serious functional flaws (`disconnect` logic not implemented, `dispatch_uid` made unique for the wrong model). It can't be accepted without a quite significant rework.

src/auditlog/migrations/0008_action_choices.py
rawad1989: 👍

PR #251, opened 2020-06-06
Fixing six module imports

PedroFagundes:
This refers to issue #231 where some imports have been moved to six directly


jjkester:
Python 2 support has been dropped in `master`, released as `1.0a1`.

PR #252, opened 2020-07-01
tests:  Check cases when reverse related field DoesNotExist.

rrauenza:
See https://github.com/jjkester/django-auditlog/issues/17

### DIFF
--- a/auditlog_tests/models.py
+++ b/auditlog_tests/models.py
@@ -60,12 +60,21 @@ class ProxyModel(SimpleModel):
         proxy = True
 
 
-class RelatedModel(models.Model):
+class RelatedModelParent(models.Model):
+    """
+    Use multi table inheritance to make a OneToOneRel field
+    """
+
+
+class RelatedModel(RelatedModelParent):
     """
     A model with a foreign key.
     """
 
-    related = models.ForeignKey(to="self", on_delete=models.CASCADE)
+    related = models.ForeignKey(to="SimpleModel", on_delete=models.CASCADE)
+    one_to_one = models.OneToOneField(
+        to="SimpleModel", on_delete=models.CASCADE, related_name="reverse_one_to_one"
+    )
 
     history = AuditlogHistoryField()
 
--- a/auditlog_tests/tests.py
+++ b/auditlog_tests/tests.py
@@ -927,7 +927,7 @@ class RegisterModelSettingsTest(TestCase):
 
         self.assertTrue(self.test_auditlog.contains(SimpleExcludeModel))
         self.assertTrue(self.test_auditlog.contains(ChoicesFieldModel))
-        self.assertEqual(len(self.test_auditlog.get_models()), 17)
+        self.assertEqual(len(self.test_auditlog.get_models()), 18)
 
     def test_register_models_register_model_with_attrs(self):
         self.test_auditlog._register_models(
@@ -1276,7 +1276,39 @@ class JSONModelTest(TestCase):
 
 
 class ModelInstanceDiffTest(TestCase):
-    def test_when_field_doesnt_exit(self):
+    def test_diff_models_with_related_fields(self):
+        """No error is raised when comparing models with related fields."""
+
+        # This tests that track_field() does indeed ignore related fields.
+
+        # a model without reverse relations
+        simple1 = SimpleModel()
+        simple1.save()
+
+        # a model with reverse relations
+        simple2 = SimpleModel()
+        simple2.save()
+        related = RelatedModel(related=simple2, one_to_one=simple2)
+        related.save()
+
+        # Demonstrate that simple1 can have DoesNotExist on reverse
+        # OneToOne relation.
+        with self.assertRaises(
+            SimpleModel.reverse_one_to_one.RelatedObjectDoesNotExist
+        ):
+            simple1.reverse_one_to_one  # equals None
+
+        # accessing relatedmodel_set won't trigger DoesNotExist.
+        self.assertEqual(simple1.relatedmodel_set.count(), 0)
+
+        # simple2 DOES have these relations
+        self.assertEqual(simple2.reverse_one_to_one, related)
+        self.assertEqual(simple2.relatedmodel_set.count(), 1)
+
+        model_instance_diff(simple2, simple1)
+        model_instance_diff(simple1, simple2)
+
+    def test_when_field_doesnt_exist(self):
         """No error is raised and the default is returned."""
         first = SimpleModel(boolean=True)
         second = SimpleModel()


jjkester:
Thanks for the fix, I restarted the Travis CI job. Let's see if everything is green.

jjkester:
Can you rebase on `master`?

rrauenza:
Sorry --just saw this recently.  Rebased and repushed.

rrauenza:
@hramezani Ok, looks ready to go!

hramezani:
Thanks @rrauenza for refreshing the patch.
If I apply your changes on `auditlog_tests/models.py` locally and then run the test, they will pass successfully.
You need to add a proper regression test for this patch.
Also, please add an entry to [changelog](https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md).

hramezani:
@rrauenza Could you please create a small project with the problem and send it to me (`hasan.r67@gmail.com`)? Then I will take a look and try to add a proper test for the patch.

Or even a model with a fixture that reproduces the problem helps as well.

Thanks!

rrauenza:
I submitted this 2 years ago, so this isn't even something I'm actively working on anymore and I'm no longer using this package.  Through a discussion with @jjkester at the time in one of the linked issues, he was trying to improve some related model stuff.  So I submitted a patch.

The code I've changed shows what the problem is -- that related field metadata doesn't have the `default ` attribute to lookup.  So we need to handle that case by checking if the field exists.

If you can find a way to trigger that code with a model than has a OneToOne related field in it, then it should reproduce.  For me, I think it was when I was importing fixtures.

rrauenza:
Hmm... this is in the diff code, even though I was having trouble with fixtures.  So maybe it can be reproduced simply in the diff tests... checking.

rrauenza:
@hramezani I'm not sure what those errors are... they kind of look unrelated?

hramezani:
@rrauenza Thanks for your information. but this error is still not reproducible for me. it means if we add your changes in [auditlog_tests/models.py](https://github.com/jazzband/django-auditlog/pull/252/files#diff-4f8b9df1dbfff67d8ddf0926cbc63c9da009e6cecfa6f373e1cd5dc4ac9cb698) and [auditlog_tests/tests.py](https://github.com/jazzband/django-auditlog/pull/252/files#diff-eb9581b1c35dba2bdbfcf71d8c389814416cdbe08a5dadadc0ce3ab4520b61b2) to the master and run the test it will be passed successfully.

> @hramezani I'm not sure what those errors are in the ci... they kind of look unrelated?

The CI failure is related to the minimum Postgres required by Django main that is updated to Postgres 12. it is already fixed in https://github.com/jazzband/django-auditlog/commit/d5d65cae2af8025b433665d3091c0b1745bc44d5. You just need to rebase your branch

rrauenza:
...it seems my test didn't get pushed?  Or when I rebased it disappeared.def 

```
test_diff_models_with_related_fields(self):                             
          """No error is raised when comparing models with related fields."""     
          first = RelatedModel()                                                  
          second = RelatedModel()                                                 
          model_instance_diff(first, second)     


```

Pushing again ... I don't have postgres locally to test against, so I need to use your cicd for testing.

rrauenza:
I think I'm narrowing this down. 

The model_instance_diff is passed with null, instance during a fixture load:

```
Based on this stack trace when loading fixtures:                      
          # virtualenv-3.6.8/lib/python3.6/site-packages/django/core/management/commands/loaddata.py line 114 in loaddata
          # self.load_label(fixture_label)                                        
          # virtualenv-3.6.8/lib/python3.6/site-packages/django/core/management/commands/loaddata.py line 181 in load_label
          # obj.save(using=self.using)                                            
          # virtualenv-3.6.8/lib/python3.6/site-packages/django/core/serializers/base.py line 223 in save
          # models.Model.save_base(self.object, using=using, raw=True, **kwargs)  
          # virtualenv-3.6.8/lib/python3.6/site-packages/django/db/models/base.py line 790 in save_base
          # update_fields=update_fields, raw=raw, using=using,                    
          # virtualenv-3.6.8/lib/python3.6/site-packages/django/dispatch/dispatcher.py line 175 in send
          # for receiver in self._live_receivers(sender)                          
          # virtualenv-3.6.8/lib/python3.6/site-packages/django/dispatch/dispatcher.py line 175 in <listcomp>
          # for receiver in self._live_receivers(sender)                          
          # virtualenv-3.6.8/lib/python3.6/site-packages/auditlog/receivers.py line 16 in log_create
          # changes = model_instance_diff(None, instance) <------- HERE           
          # virtualenv-3.6.8/lib/python3.6/site-packages/auditlog/diff.py line 135 in model_instance_diff
          # new_value = get_field_value(new, field)                               
          # virtualenv-3.6.8/lib/python3.6/site-packages/auditlog/diff.py line 80 in get_field_value
          # value = field.default if field.default is not NOT_PROVIDED else None  
          # AttributeError: Problem installing fixture 'data.json': 'OneToOneRel' object has no attribute 'default'
```

For this to happen, though, we must trigger an ObjectDoesNotExist.

(code from before my change:)
```
try:                                                                        
          if isinstance(field, DateTimeField):                                    
              # DateTimeFields are timezone-aware, so we need to convert the field
              # to its naive form before we can accurately compare them for changes.
              if value is not None and settings.USE_TZ and not timezone.is_naive(value):
          elif isinstance(field, JSONField):                                      
          else:                                                                   
              value = smart_str(getattr(obj, field.name, None))                   
      except ObjectDoesNotExist:                                                  
```

I'm not sure how to trigger this.  I've seen it when model.foo is a FK that points to None.  Oh.. I think its triggered from the reverse lookup from the fk model. 

Trying more tests...

rrauenza:
FYI, still pushing tests -- we actually want them to fail as I'm trying to test that we catch the failure case -- fix is disabled currently.

hramezani:
I think it's better to run the tests locally. Because running them on CI takes time

rrauenza:
@hramezani I think I have figured out what is going on now.

I think since my proposed change long ago, I think the code now filters what fields go through the diff generator.  In the past, the reverse related field got captured.  Which actually doesn't make sense to do anyway.  So I suspect somewhere we only grab "local" fields to the object.  So the code is never going to trigger because the one to one related field type never gets to the diff function.

What I'd like to do is add a unit test for this, without the fix.

rrauenza:
In fact, it looks like this is where we avoid remote fields::

```
def track_field(field):                                                         
      """                                                                         
      Returns whether the given field should be tracked by Auditlog.              
                                                                                  
      Untracked fields are many-to-many relations and relations to the Auditlog LogEntry model.
                                                                                  
      :param field: The field to check.                                           
      :type field: Field                                                          
      :return: Whether the given field should be tracked.                         
      :rtype: bool                                                                
      """                                                                         
      from auditlog.models import LogEntry                                        
                                                                                  
      # Do not track many to many relations                                       
      if field.many_to_many:                                                      
          return False                                                            
                                                                                  
      if (                                                                        
          getattr(field, "remote_field", None) is not None                        
          and field.remote_field.model == LogEntry                                
      ):                                                                          
          return False                                                            
                                                                                  
      return True              
```

hramezani (approved):
Thanks @rrauenza for your effort :+1:

PR #256, opened 2020-08-06
fix backward compatibility issues with django 3.0

rossi1:


mirzadelic:
@rossi1 but should update python-dateutil or set >= because of other dependencies.

pcamelo (approved):

pcamelo:
@jjkester could you review this PR, thanks in advance.

src/auditlog/filters.py
```diff
@@ -1,6 +1,7 @@
 from django.contrib.admin import SimpleListFilter
 
 
+
```
jjkester, line 4: Blank line doesn't seem to be a necessary change

src/auditlog/models.py
```diff
@@ -3,15 +3,18 @@
 import json
 import ast
 
+from six import python_2_unicode_compatible
+from six import iteritems, integer_types
+
```
jjkester, line 8: These imports could be a bit neater

jjkester (changes requested):
Thanks for the concise MR! Could you have a look at the following:

- Unnecessary blank lines
- Import formatting/ordering

jjkester:
I have decided to go ahead with #257 to drop all unsupported Python versions, saves the trouble of using six.

Nevertheless, thanks for the effort of creating a PR.

src/auditlog/models.py
rossi1: I will fix this once I get on my computer

paulpavlishcoupa:
> I have decided to go ahead with #257 to drop all unsupported Python versions, saves the trouble of using six.
> 
> Nevertheless, thanks for the effort of creating a PR.

Sounds good, I'll close the PR now.

src/auditlog/models.py
rossi1: Unnecessary blank lines removed and imports have been formatted properly

jjkester:
Closed, as Django 3.0 support has been added in master. Release 1.0a1 is available. Python 2 is no longer supported, as it has been deprecated for a long time and support finally ended this January.

PR #258, opened 2020-09-01
Adding optional OpenTracing integration.

ndwhelan:
[OpenTracing](https://opentracing.io/) provides the ability for
distributed tracing independent of vendor or downstream consumer.

This commit optionally adds an OpenTracing span in the receiver methods. The
spans are defined as `django-auditlog.{action}.{model_name}`, where
`action` is either `create`, `update`, or `delete`. This allows for
analysis on the performance impacted created through audit logging.

To install it, include the extra "opentracing"
Ex: `pip install django-auditlog[opentracing]`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 django-jsonfield>=1.0.0
-python-dateutil==2.6.0
+opentracing==2.3.0
+python-dateutil==2.8.1
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from distutils.core import setup
 
 setup(
     name='django-auditlog',
-    version='0.4.7',
+    version='0.4.7.1',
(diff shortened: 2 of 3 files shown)


ndwhelan:
I am curious if this is something that could get merged in? I know that it is adding a dependency, though optional, so I understand if there is not a desire to merge this.

Best,

Nick

jjkester:
Maybe something generic, like custom signals, would be better? Then integrating opentracing (or something else) would be fairly easy for a developer, including customizing opentracing.

This library is not very opinionated, and I think it is good to keep it that way. One dev may want opentracing, another dev something different. A hook (signal) would keep options open.

ndwhelan:
Fair enough on not wanting to prescribe OpenTracing. I'm going to look at something with custom signals. I'll go ahead with closing this, and will open another PR. Thanks!

PR #259, opened 2020-09-01
Add support for using DB column in diff.

ndwhelan:
Added a new field `db_fields` to `register`. Any field specified here
will use the DB column name, instead of the field name. If a custom DB
column was specified when defining the field, that will be used.

Relations are not followed, leading to significant performance gains
when a relational field is specified in `db_fields`. For example, if a
field called `"relation"` exists and is specified in `db_fields`,
`relation_id` is used (or the custom column name) and `relation` is
excluded.

### DIFF
--- a/src/auditlog/diff.py
+++ b/src/auditlog/diff.py
@@ -1,10 +1,20 @@
 from __future__ import unicode_literals
 
+from collections import namedtuple
+
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
-from django.db.models import Model, NOT_PROVIDED, DateTimeField
+from django.db.models import DateTimeField, Model, NOT_PROVIDED
 from django.utils import timezone
 from django.utils.encoding import smart_text
+
+ImplicitField = namedtuple("Field", "name")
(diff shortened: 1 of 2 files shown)


jjkester:
I think the `db_fields` name is somewhat confusing (I half-expected to have to put strings in there directly referencing db fields.)

Maybe a nicer approach is to keep just the `include_fields` and `exclude_fields` options and let the user specify either strings (current behavior) or something explicit by using a type (e.g. `DbField`, `DbValue`, ...), similar to what your code currently already does internally. That keeps the `register` interface clean and easy to use.

Example:

```python
auditlog.register(MyModel, include_fields=['slug', DbField('user')])
```

Not sure (yet) how the logic for resolving includes/excludes should work with this.

ndwhelan:
Thanks for the feedback. I'm not really sure what I was thinking with that name, but I agree it's confusing. For some reason I thought this lined up with how you can do something similar on the Django admin panel. But I don't know why I thought that now...

I'll close this and look in to implementing it as you suggest.

Issue #260, opened 2020-09-02
Incorrectly detecting change of zeroes in decimal place


Liamhanninen:
First of all this library has been so useful. It's exactly what I needed and works really well. There is only one issue. 
It seems to be detecting changes that aren't occurring. At first I thought the model was adding zeros or the db was removing them. But after some digging I think it's a bug with django-auditlog.
It thinks a change occurred anytime I save this record: it thinks that I started with .00 and changed it to .0. But .00 is the existing value and I didn't change anything. I just saved() the record. 
![image](https://user-images.githubusercontent.com/15257781/91921283-1e41c680-ec99-11ea-98cb-08ae44b2c1e7.png)
I checked the output of the form itself in the clean() method: 
```
services_sub_total
44.00
sum_payments
33.00
```
And then in the database:
![image](https://user-images.githubusercontent.com/15257781/91921437-8c868900-ec99-11ea-95ae-eb37d8df2183.png)
I just cant figure out where/why this is detecting only one zero in the decimal place - I'm not seeing that .0 anywhere. 
Could the delta functions be truncating multiple zeros in the decimals to .0?

iklinac:
As already responded to [SO]( https://stackoverflow.com/questions/63713985/what-happens-after-django-clean-method-but-before-commit/63714543#63714543) question

If you look up into [source][1] you would see that it compares these two values not as field values but through use of [smart_text][2]  values are converted to string inside [get_field_value][3]

So it basically compares "44.00" and your value of "44.0" and this is different string so it is triggered as change

Not sure if this is by design as in my eyes would be considered bug




  [1]: https://github.com/jjkester/django-auditlog/blob/master/src/auditlog/diff.py#L137
  [2]: https://docs.djangoproject.com/en/3.1/ref/utils/#django.utils.encoding.smart_text
  [3]: https://github.com/jjkester/django-auditlog/blob/master/src/auditlog/diff.py#L78

Liamhanninen:
Thanks for digging into that and providing the file and line! I will hack it for now. Hopefully it is fixed here (I'm sure my way will not be the best).

PR #261, opened 2020-09-06
Fix for Issue #260

Liamhanninen:
This is my first pull request. I hope this can help. Please let me know if you have any questions or better ideas on how to fix this.

EXAMPLE ISSUE: The field's decimal places = 2. So the db stores .00. But in the model's save() method performs calculations on that field which result in python truncating the zeros to .0. This causes model_instance_diff() to incorrectly detect a change.

THE FIX:  check if the field is a DecimalField. If so then remove all trailing zeros and the decimal. Used the code from this answer https://stackoverflow.com/questions/11227620/drop-trailing-zeros-from-decimal

### DIFF
--- a/src/auditlog/diff.py
+++ b/src/auditlog/diff.py
@@ -78,7 +78,8 @@ def get_field_value(obj, field):
         # DecimalFields can be stored with multiple zeroes after decimal but 
         # might only have one zero after save() method, so we need to remove 
         # all extra zeroes for an accurate comparison.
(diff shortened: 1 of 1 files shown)


jjkester:
I think this can be resolved in a smarter way, by finding an alternative to the string conversion that currently happens.

Most likely comparing bare values works, but this is something that may not be the case for all values. This way we can leverage the `__eq__` magic method of Python.

Creating special cases for fields is something I rather not do, as you are likely to end up with many of them that all need to be maintained.

Issue #266, opened 2020-09-23
v1.0a1 serializes additional_data log entry field as string


l0b0:
After upgrading django-auditlog to v1.0a1 I now have to `json.loads(log_entry["additional_data"])` to get the same result as `log_entry["additional_data"]` before. Is this an intentional change?

Real-Gecko:
I created a PR https://github.com/jazzband/django-auditlog/pull/271 which replaces `JSONField` with `TextField` while retaining original functionality. It passes all tox tests, try it out.

l0b0:
That would be very, very bad for the project I'm on. We already have many millions of audit log entries. Why not leave it as a `JSONField` and deserialize it as before?

Real-Gecko:
I don't know what changed in Django 3.1, but it fails tests with JSONField, replacing it with TextField solves the problem.

> We already have many millions of audit log entries.

This won't destroy existing data, there're two migrations: first alters field type, second renames it to `additional_data_json`. In case you're not using PostgreSQL there even won't be type conversion on the column:
https://github.com/adamchainz/django-jsonfield/blob/a4a5f339df0da850a1087ae7b3cd20a19ef8c6b8/jsonfield/fields.py#L71-L81
And you can still get `log_entry["additional_data"]`, so I think no changes will be required in code.
However if you're tied to the field name of `additional_data` somewhere outside Django this will cause trouble.

l0b0:
JSONField is built into Django 3, you should be able to use that directly now.

We already had to `loads(log_entry["additional_data"]` when upgrading to 1.0a1, are you saying that's meant to stay that way? Why change it? Is `additional_data` *meant* to be freeform text rather than JSON?

The comment about millions of records wasn't about losing them, but rather the migration taking possibly a very long time on a production system. I still don't quite understand why this shouldn't just stay a JSONField, simplifying the code, avoiding any JSON encoding/decoding incompatibilities, and saving time when upgrading.

Real-Gecko:
> We already had to `loads(log_entry["additional_data"]` when upgrading to 1.0a1

Nope, it's exactly what your issue about, `log_entry.additional_data` is Python dict.

This two methods ensure that it'll be valid:
```python

    @property
    def additional_data(self):
        data = None
        try:
            data = json.loads(self.additional_data_json)
        except:
            pass
        return data

    @additional_data.setter
    def additional_data(self, var):
        self.additional_data_json = json.dumps(var)
```

> shouldn't just stay a JSONField

[...]

PR #271, opened 2020-09-28
Replaced JSONField with TextField

Real-Gecko:


ndwhelan:
This seems like a downgrade to me. It doesn't seem like the best response to issue #266

auditlog/models.py
```diff
@@ -305,6 +305,19 @@ def changes_display_dict(self):
...
         return changes_display_dict
 
+    @property
+    def additional_data(self):
+        data = None
+        try:
+            data = json.loads(self.additional_data_json)
+        except:
+            pass
+        return data
```
jezdez, line 315: This shouldn't be a blank exeption handling but instead checking for `json.JSONDecodeError` and return a default value of `{}`.
```suggestion
        try:
            return json.loads(self.additional_data_json)
        except json.JSONDecodeError:
            return {}
```

auditlog/models.py
```diff
@@ -179,8 +178,9 @@ class Action:
                               related_name='+', verbose_name=_("actor"))
     remote_addr = models.GenericIPAddressField(blank=True, null=True, verbose_name=_("remote address"))
     timestamp = models.DateTimeField(auto_now_add=True, verbose_name=_("timestamp"))
-    additional_data = JSONField(blank=True, null=True, verbose_name=_("additional data"))
-
+    additional_data_json = models.TextField(blank=True,
```
jezdez, line 181: I'm not sure if this is the right name for the column to be honest, since it's infact not a JSON field anymore but just plain text (since there is a real JSONField in this new Django era). 

How about just naming it like that (and removing the rename migration and updating the model property)?
```suggestion
    additional_data_text = models.TextField(blank=True,
```

jezdez (changes requested):
I would argue that this is a sensible work-around for the situation that django-auditlog finds itself in now, an era where there is an actual JSONField in Django instead of the pseudo JSONField from django-jsonfield.

`additional_data` hasn't really been used much in the app I think, so I don't think we should over-optimize by switching to Django's JSONField but treat it as a dumb field to store data in.

Of course alternatively we could switch to [django-jsonfield-backport](https://pypi.org/project/django-jsonfield-backport/) and actually convert all the data from the TextField to an honest JSON field but that would be a pretty big data migration for most installations given the purpose of the app.

Linkid:
I think it could be useful to switch to [django-jsonfield-backport](https://pypi.org/project/django-jsonfield-backport/) for older versions than 3.1 (e.g. 2.2).

DanielSchaffer:
If the specific implementation of the field isn't important to this library, it'd be nice to avoid adding any package requirement just for the field. I'd _really_ like to use this for my project's audit logging, but we're in a bit of a complicated spot - we're on 2.2, getting ready to upgrade to 3.2, and we already use the `JSONField` from `django_mysql` everywhere else we use JSON. I'm really reluctant to add another `JSONField` that could potentially complicate things for us. It'd be awesome if the field class could be customized, with whatever default you land on set up as an optional dependency.

Linkid:
Django 2.2 will be unmaintained after April 2022. After that date, there will be no more dependency for the `JSONField` field.
In #339, the `django-jsonfield` dependency was replaced by `django-jsonfield-backport` as suggested. So, I think we can close this PR for now.
Thanks @Real-Gecko for this PR anyway.

PR #273, opened 2020-09-29
Update models.py

viktor2097:
Added a check if choices attribute exists, otherwise it might throw an error like: AttributeError: 'ManyToOneRel' object has no attribute 'choices'

### DIFF
--- a/auditlog/models.py
+++ b/auditlog/models.py
@@ -258,7 +258,7 @@ class LogEntry(models.Model):
             values_display = []
             # handle choices fields and Postgres ArrayField to get human readable version
             choices_dict = None
-            if getattr(field, 'choices') and len(field.choices) > 0:
+            if getattr(field, 'choices', None) and len(field.choices) > 0:
                 choices_dict = dict(field.choices)
             if hasattr(field, 'base_field') and isinstance(field.base_field, Field) and getattr(field.base_field, 'choices') and len(field.base_field.choices) > 0:
                 choices_dict = dict(field.base_field.choices)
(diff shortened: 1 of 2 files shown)


auditlog/models.py
```diff
@@ -258,7 +258,7 @@ def changes_display_dict(self):
             values_display = []
             # handle choices fields and Postgres ArrayField to get human readable version
             choices_dict = None
-            if getattr(field, 'choices') and len(field.choices) > 0:
+            if getattr(field, 'choices', None) and len(field.choices) > 0:
```
ndwhelan, line 261: There's not a case where this doesn't exist, unless you go so far as to consider some crazy monkey patching or other questionable practices, in which I suspect many more things will break as well. `choices` is set in `__init__()`` of `django.db.models.fields.Field`

ndwhelan (dismissed):

auditlog/models.py
viktor2097: It's cascading from:

> django          |   File "/usr/local/lib/python3.8/site-packages/django/template/base.py", line 828, in _resolve_lookup
> django          |     current = current[bit]
> django          | TypeError: 'LogEntry' object is not subscriptable

And by model is defined as defined in docs

> history = AuditlogHistoryField()
> auditlog.register(Invoice)

Considering LogEntry is from the library itself, it's surely a bug in this package.
I have even gone so far to delete all fields except a simple CharField and LogEntry still cascades the .choices error. If this is a bug related to the django 3.x update or not, i don't know.
I won't investigate any further considering this commit solved the pain points of changes_display_dict

Issue #277, opened 2020-10-16
Remote Address not set if user not authenticated


ndwhelan:
When the user is not authenticated, the remote address is not added to the audit log, even if it is known.

`auditlog.middleware.set_actor` should be set as a `pre_save` signal receiver, even if the user is not authenticated. Furthermore, deferring the authentication check until that signal receiver allow recording the actor, even if it is set outside of middleware, or in a following middleware.

PR #278, opened 2020-10-16
Moving user auth check in to the `set_actor()`

ndwhelan:
Move the check for whether the user is authenticated from before
registering `set_actor()` as the `pre_save` signal receiver
in to `set_actor()`. This enables two things:

1. `remote_addr` in the `LogEntry` is set even if the user is
    not authenticated.
2. This middleware no longer execute before the user is set
   on the request. Some frameworks may set this in a
   following middleware. Some frameworks, such as the widely
   adopeted Django Rest Framework do not set this as part
   of a middleware. So, the upshot of hit is that the
   actor will not be set when using Django Rest Framework.


ndwhelan:
Solves #277

PR #262, opened 2020-09-07
Add set_actor conext manager

alieh-rymasheuski:
Fixes https://github.com/jjkester/django-auditlog/issues/211

This pull request moves the logic to install and remove listeners for Django signals to a context manager which is now used from the middleware, but also can be used in non-request context.

Example usage (for the use case of celery tasks):
```python
from auditlog.context import set_actor

@app.task
def do_stuff(actor_id: int):
    actor = get_user(actor_id)
    with set_actor(actor):
        # if your code here leads to creation of LogEntry instances, these will have actor set
        ...
```

PR #279, opened 2020-10-16
Moving middleware user auth check to `set_actor()`

ndwhelan:
Move the check for whether the user is authenticated from before
registering `set_actor()` as the `pre_save` signal receiver
in to `set_actor()`. This enables two things:

1. `remote_addr` in the `LogEntry` is set even if the user is
   not authenticated.
2. This middleware no longer execute before the user is set
   on the request. Some frameworks may set this in a
   following middleware. Some frameworks, such as the widely
   adopeted Django Rest Framework do not set this as part
   of a middleware. So, the upshot of hit is that the
   actor will not be set when using Django Rest Framework.

### DIFF
--- a/auditlog/middleware.py
+++ b/auditlog/middleware.py
@@ -34,7 +34,7 @@ class AuditlogMiddleware(MiddlewareMixin):
             threadlocal.auditlog['remote_addr'] = request.META.get('HTTP_X_FORWARDED_FOR').split(',')[0]
 
         # Connect signal for automatic logging
-        if hasattr(request, 'user') and getattr(request.user, 'is_authenticated', False):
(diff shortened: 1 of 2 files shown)


ndwhelan:
Addresses #277

ndwhelan:
This is probably solved by PR #262

alieh-rymasheuski:
This is actually not solved in #262. #262 sets both actor and remote_addr only for authenticated users, which is exactly the issue attempted to be resolved by this pull request.

aqeelat:
Superseded by #484
@hramezani

hramezani:
Thanks @ndwhelan for the initial patch and @aqeelat for continuing on this.

PR #280, opened 2020-10-17
Adding Foriegn Key Only Fields

ndwhelan:
When building the changes JSON for a LogEntry, following
relations (e.g. `ForeigKey` fields) can resutl in a
significant performance penalty. This commit adds support for
only logging to foreign key value, by specifying
fields to log in this manner wiht a `fk_only_fields`
when registering a model with the audit log.


aqeelat:
Thank you @ndwhelan for your PR. Can you please rebase it?

PR #281, opened 2020-10-23
Simplified travis config and added Jazzband release config.

jezdez:
This relates to #268.


jezdez:
@jjkester To reply to your question in #268, it's better to follow the Jazzband best practice to deploy on tag creation to the Jazzband package index and then use it to review/confirm the deploy to PyPI.

I've simplified a few things here that should make things easier for moving forward, like using tox-travis and migrating to travis-ci.com (.org will be shut down at the end of the year).

I do notice that there is a test currently breaking on Django 3.1 due to differences in how the additional data field is handled differently in django-jsonfield (older versions returning a dict, Django 3.1 returning a string).

Please let me know how we can move forward with this, I've pushed the branch to the repo, so feel free to append my changes and/or ask questions. Thank you!

PR #282, opened 2020-11-10
Ignore null changeset on creation and deletion

joeyjoejoejr:
## Problem
I registered a nullable field of a model using `include_fields`.  There was a a check for null `changes` in the `log_update` receiver, but not the `log_create` and `log_delete` ones.  This meant that a log entry was created with `null` for the changes field whenever a model was created or updated, but didn't have a value for that field.

When the mixin method `msg_short` is called in the Audit Log admin index, it would error because it's expecting an object for changes.

## What Changed:
* Made the SimpleIncludeModel label field nullable
* Added tests for creation and deletion of a model with a null label field
* Added checks to `log_create` and `log_delete` to make sure entries aren't created if the changes are None.

### DIFF
--- a/auditlog/receivers.py
+++ b/auditlog/receivers.py
@@ -13,11 +13,13 @@ def log_create(sender, instance, created, **kwargs):
     if created:
         changes = model_instance_diff(None, instance)
 
-        log_entry = LogEntry.objects.log_create(
-            instance,
-            action=LogEntry.Action.CREATE,
-            changes=json.dumps(changes),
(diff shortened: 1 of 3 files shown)


hramezani:
Thanks @joeyjoejoejr for the patch :+1: 
Could you please rebase your patch?

aqeelat:
@hramezani I think will be solved as a side-effect of #483. We might still need the tests but let's evaluate after we merge #483.

PR #283, opened 2020-11-25
Initial GitHub Actions workflow.

jezdez:
Travis CI has a new pricing model which places limits on open source.

* https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
* https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

Many projects are moving to GitHub Actions instead, including Jazzband projects: 

* https://github.com/orgs/jazzband/projects/1
* https://github.com/jazzband-roadies/help/issues/206

This is based on https://github.com/jazzband/contextlib2/pull/26.

This contains a work-around for #266 until there is a real fix, e.g. via #271 etc.

No more mention of Travis CI in the codebase:

```console
$ git grep -i travis
$
```


.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
+name: Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
```
hugovk, line 8: This would be useful:
```suggestion
    strategy:
      fail-fast: false
```

At the moment, if a single job fails, the rest are all auto-cancelled.

This can be especially annoying for temporary time-outs. It was also a pain for my other PR: first 3.3 failed because it wasn't supported, which cancelled the 2.7 job, and masking that I also needed to pin dependencies for 2.7. 

Turning it off allows the other jobs to finish running, which can help you determine if the failure is a compatibility bug (e.g. fails on 2.7, passes on all 3.x) or a more common coding bug (fails on all).

The default auto-cancelling behaviour can be useful, but I find best for when there's a particularly large and slow-running matrix for a busy repo.

Otherwise for quick builds, like most of these (e.g. only 2 minutes here), I suggest letting them all run.

auditlog_tests/tests.py
```diff
@@ -257,9 +258,13 @@ def test_model_with_additional_data(self):
         self.assertTrue(obj_with_additional_data.history.count() == 1,
                         msg="There is 1 log entry")
         log_entry = obj_with_additional_data.history.get()
-        self.assertIsNotNone(log_entry.additional_data)
-        extra_data = log_entry.additional_data
-        print(type(extra_data), extra_data)
+        # FIXME: Work-around for the fact that additional_data isn't working
+        # on Djang 3.1 correct (see https://github.com/jazzband/django-auditlog/issues/266)
```
hugovk, line 262:
```suggestion
        # on Django 3.1 correctly (see https://github.com/jazzband/django-auditlog/issues/266)
```

hugovk (commented):
Looks good, just a few suggestions!

hugovk (approved):

hugovk:
I spotted the cron was still enabled at https://travis-ci.com/github/jazzband/django-auditlog / https://travis-ci.com/github/jazzband/django-auditlog/settings:

image

So I deleted it:

image

jezdez:
Thanks!

PR #288, opened 2020-12-06
Qa

hramezani:


tox.ini
```diff
@@ -35,6 +36,15 @@ changedir = docs/source
...
 commands = sphinx-build -W -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
 
+[testenv:py38-qa]
+basepython = python3.8
+deps =
+    black
+    isort
+commands =
+    black -t py36 --check --diff auditlog auditlog_tests setup.py runtests.py
+    isort --check-only --diff --profile black auditlog auditlog_tests setup.py runtests.py
```
smithdc1, line 46: I think it would be good to add the isort profile as a setting generally to the project. This makes it easier when folk want to apply isort as they don't need to pass it in as a cli option.

tox.ini
```diff
@@ -35,6 +36,15 @@ changedir = docs/source
 deps = -rdocs/requirements.txt
 commands = sphinx-build -W -b html -d {envtmpdir}/doctrees . {envtmpdir}/html
 
+[testenv:py38-qa]
+basepython = python3.8
+deps =
+    black
+    isort
+commands =
+    black -t py36 --check --diff auditlog auditlog_tests setup.py runtests.py
```
smithdc1, line 45: Py35 is still in the test matrix?

tox.ini
hramezani: Thanks :+1: . I created a `project.tmol` file and move both `black` and `isort` configuration there.

tox.ini
hramezani: I changed it to `py35` in `project.toml` file.

setup.py
```diff
@@ -7,29 +7,31 @@
...
-        'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.4',
-        'Programming Language :: Python :: 3.5',
-        'Programming Language :: Python :: 3.6',
-        'Programming Language :: Python :: 3.7',
-        'Programming Language :: Python :: 3.8',
-        'Programming Language :: Python :: 3.9',
-        'License :: OSI Approved :: MIT License',
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.4",
```
smithdc1, line 29: Side note, Python 3.4 needs removing.

smithdc1 (approved):
LGTM

setup.py
hramezani: Good catch :+1: . Remove!

smithdc1 (approved):

Issue #267, opened 2020-09-23
RemovedInDjango40Warning: smart_text() is deprecated in favor of smart_str()


l0b0:
This warning comes up a lot while using v1.0a1 with Django 3.1.1. Just a heads up.

Linkid:
Done in #295

PR #295, opened 2021-03-10
Replace deprecated smart_text() with smart_str()

sp1thas:
The purpose of this PR is to solve #267 by replacing `django.utils.encoding.smart_text` with `django.utils.encoding.smart_str` whenever is applicable (Django >= 3.0)

### DIFF
--- a/auditlog/diff.py
+++ b/auditlog/diff.py
@@ -2,6 +2,7 @@ from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import NOT_PROVIDED, DateTimeField, Model
 from django.utils import timezone
+
 try:
     from django.utils.encoding import smart_str
 except ImportError:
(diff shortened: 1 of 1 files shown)


auditlog/diff.py
```diff
@@ -2,7 +2,11 @@
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import NOT_PROVIDED, DateTimeField, Model
 from django.utils import timezone
-from django.utils.encoding import smart_text
+
+try:
+    from django.utils.encoding import smart_str
+except ImportError:
+    from django.utils.encoding import smart_text as smart_str
```
hramezani, line 9: Why you don't import `smart_str` directly like:

```suggestion
from django.utils.encoding import smart_text
```

hramezani:
@sp1thas Thanks for the PR. I include your change into https://github.com/jazzband/django-auditlog/pull/302.

PR #303, opened 2021-05-25
Add django main to test pipeline.

hramezani:


auditlog/migrations/0004_logentry_detailed_object_repr.py
```diff
@@ -1,5 +1,10 @@
-import jsonfield.fields
 from django.db import migrations, models
+from django.utils.version import get_complete_version
+
+if get_complete_version() >= (3, 1):
+    from django.db.models import JSONField
+else:
+    from jsonfield.fields import JSONField
```
alieh-rymasheuski, line 7: This looks a dangerous change to me. Does this mean that if I initialize my database on Django 3.1, I'll get native JSON field, but if I initialize my database on Django 3.0 and then upgrade to Django 3.1, I'll get a char field storing JSON data? This change to an old migration doesn't feel right to me.

alieh-rymasheuski (changes requested):

auditlog/migrations/0004_logentry_detailed_object_repr.py
hramezani: hmmm, I haven't considered this case. I will close it.

PR #308, opened 2021-06-28
Pin psycopg2 to version compatible with Django 2.2

alieh-rymasheuski:
See https://github.com/psycopg/psycopg2/issues/1293#issuecomment-862835147. Without this change, all CI tests fail on the current master branch.

PR #309, opened 2021-06-28
Add logic to track changes to m2m fields

alieh-rymasheuski:
# Links

This branch is based on https://github.com/jazzband/django-auditlog/pull/308 for the tests to work.

This pull request started as an attempt to port https://github.com/jazzband/django-auditlog/pull/249 to my fork of django-auditlog, but a big portion of it was rewritten in the process.

# Usage

```python
auditlog.register(User, m2m_fields={"tags", "contacts"})
```

# Example

image

# Solution overview

When the user sets `m2m_fields` when registering the model, we generate a signal handler for each of the listed fields (see `make_log_m2m_changes`) and (a) subscribe it to `m2m_changed` signal sent by the `through` model and (b) store it in a mapping which can be subsequently used to undo this subscription.

The m2m events are marked with `UPDATE` action and can be identified through a new format of the value in `changes` field:
```js
{
  "<normal field>": ["<old value>", "<new value>"],
  "<m2m field>": {
    "type": "m2m",  // trying to be future-proof here
    "operation": "add",  // or "delete"
    "objects": ["string representation", "of each", "affected object"],
  },
}
```

Admin representation logic has been updated to match the new format.

Note that when the user changes multiple m2m fields on the same object through the admin, both adding and removing some objects from each, this code will generate multiple log entries: each log entry will represent a single operation (add or delete) of a single field (e.g. if you both add and delete values from 2 fields on the same form in the same request, you'll get 4 log entries).

### DIFF
--- a/auditlog/models.py
+++ b/auditlog/models.py
@@ -76,6 +76,55 @@ class LogEntryManager(models.Manager):
             )
         return None
 
+    def log_m2m_changes(
+        self, changed_queryset, instance, operation, field_name, **kwargs
+    ):
+        """Create a new "changed" log entry from m2m record.
+
+        :param instance: The model instance to log a change for.
+        :type instance: Model
+        :param operation: "add" or "delete".
+        :type action: str
+        :param kwargs: Field overrides for the :py:class:`LogEntry` object.
+        :return: The new log entry or `None` if there were no changes.
+        :rtype: LogEntry
+        """
+
+        pk = self._get_pk_value(instance)
+        if changed_queryset is not None:
+            kwargs.setdefault(
+                "content_type", ContentType.objects.get_for_model(instance)
+            )
+            kwargs.setdefault("object_pk", pk)
+            kwargs.setdefault("object_repr", smart_str(instance))
+            kwargs.setdefault("action", LogEntry.Action.UPDATE)
+
+            if isinstance(pk, int):
+                kwargs.setdefault("object_id", pk)
+
+            get_additional_data = getattr(instance, "get_additional_data", None)
+            if callable(get_additional_data):
+                kwargs.setdefault("additional_data", get_additional_data())
+
+            objects = [smart_str(instance) for instance in changed_queryset]
+            kwargs["changes"] = json.dumps(
+                {
+                    field_name: {
+                        "type": "m2m",
+                        "operation": operation,
+                        "objects": objects,
+                    }
+                }
+            )
+            db = instance._state.db
+            return (
+                self.create(**kwargs)
+                if db is None or db == ""
+                else self.using(db).create(**kwargs)
+            )
+
+        return None
+
     def get_for_object(self, instance):
         """
         Get log entries for the specified model instance.
--- a/auditlog_tests/tests.py
+++ b/auditlog_tests/tests.py
@@ -21,8 +21,10 @@ from auditlog_tests.models import (
     CharfieldTextfieldModel,
     ChoicesFieldModel,
     DateTimeFieldModel,
+    FirstManyRelatedModel,
     ManyRelatedModel,
     NoDeleteHistoryModel,
+    OtherManyRelatedModel,
     PostgresArrayFieldModel,
     ProxyModel,
     RelatedModel,
@@ -142,7 +144,7 @@ class ProxyModelTest(SimpleModelTest):
 
 class ManyRelatedModelTest(TestCase):
     """
-    Test the behaviour of a many-to-many relationship.
+    Test the behaviour of a default many-to-many relationship.
     """
 
     def setUp(self):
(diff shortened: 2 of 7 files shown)


doom9ra:
Do these changes include reverse FK tracking?

naveenboobalakrishnan:
Any idea when this PR will be merged with prod? Would like to use this in our production system.

auditlog/models.py
```diff
@@ -69,6 +69,50 @@ def log_create(self, instance, **kwargs):
             return self.create(**kwargs)
         return None
 
+    def log_m2m_changes(
+        self, changed_queryset, instance, operation, field_name, **kwargs
+    ):
+        """Create a new "changed" log entry from m2m record.
+
```
hramezani, line 76: I think you forgot to cover all params here

auditlog/models.py
```diff
@@ -69,6 +69,50 @@ def log_create(self, instance, **kwargs):
...
+            kwargs.setdefault("object_pk", pk)
+            kwargs.setdefault("object_repr", smart_str(instance))
+            kwargs.setdefault("action", LogEntry.Action.UPDATE)
+
+            if isinstance(pk, int):
+                kwargs.setdefault("object_id", pk)
+
+            get_additional_data = getattr(instance, "get_additional_data", None)
+            if callable(get_additional_data):
+                kwargs.setdefault("additional_data", get_additional_data())
```
hramezani, line 100: This line doesn't have test coverage.

auditlog/models.py
alieh-rymasheuski: You're right, thank you.

hramezani:
Please add `m2m_fields` to `AUDITLOG_INCLUDE_TRACKING_MODELS` example in https://github.com/jazzband/django-auditlog/blob/master/docs/source/usage.rst as well.

Also, please update this test in include `m2m_attrs` in the test.
https://github.com/jazzband/django-auditlog/blob/bcd0d43566f24a08839ac17e85864ad7b7ae8a90/auditlog_tests/tests.py#L932

auditlog/models.py
alieh-rymasheuski: Fixed.

alieh-rymasheuski:
@hramezani, all done. I reworked the tests for ManyRelatedModel while working on the improvements you requested in addition to the extra documentation.

auditlog_tests/tests.py
```diff
@@ -947,6 +989,21 @@ def test_register_models_register_model_with_attrs(self):
         self.assertEqual(fields["include_fields"], ["label"])
         self.assertEqual(fields["exclude_fields"], ["text"])
 
+    def test_register_models_register_model_with_attrs__m2m_fields(self):
```
hramezani, line 992:
```suggestion
    def test_register_models_register_model_with_attrs_m2m_fields(self):
```

docs/source/usage.rst
```diff
@@ -91,6 +93,17 @@ For example, to mask the field ``address``, use::
 
     Masking fields
 
+**Many-to-many fields**
+
+.. versionadded:: 2.1.0
```
hramezani, line 98: Please move it to the end of this section

docs/source/usage.rst
```diff
@@ -91,6 +93,17 @@ For example, to mask the field ``address``, use::
...
+
+Changes to many-to-many fields are not tracked by default. If you want to enable tracking if a many-to-many field on a model, pass ``m2m_fields`` to the ``register`` method::
+
+    auditlog.register(MyModel, m2m_fields={"tags", "contacts"})
```
hramezani, line 102:
```suggestion
Changes to many-to-many fields are not tracked by default. If you want to enable tracking if a many-to-many field on a model, pass ``m2m_fields`` to the ``register`` method:

.. code-block:: python

    auditlog.register(MyModel, m2m_fields={"tags", "contacts"})
```

docs/source/usage.rst
```diff
@@ -91,6 +93,19 @@ For example, to mask the field ``address``, use::
 
     Masking fields
 
+**Many-to-many fields**
+
+Changes to many-to-many fields are not tracked by default. If you want to enable tracking if a many-to-many field on a model, pass ``m2m_fields`` to the ``register`` method::
```
hramezani, line 98:
```suggestion
Changes to many-to-many fields are not tracked by default. If you want to enable tracking of a many-to-many field on a model, pass ``m2m_fields`` to the ``register`` method:
```

alieh-rymasheuski: Fixed.

hramezani (approved):

hramezani:
Great work @alieh-rymasheuski :+1:

PR #333, opened 2021-12-08
use TRUNCATE TABLE in auditlogflush command

MHM5000:
- use TRUNCATE TABLE in auditlogflush command

### DIFF
--- a/auditlog/management/commands/auditlogflush.py
+++ b/auditlog/management/commands/auditlogflush.py
@@ -1,5 +1,5 @@
 from django.core.management.base import BaseCommand
-from django.db import connections
+from django.db import connection
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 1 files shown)


auditlog/management/commands/auditlogflush.py
```diff
@@ -28,7 +27,8 @@ def handle(self, *args, **options):
             answer = response == "y"
 
         if answer:
-            count, _ = LogEntry.objects.all().delete()
-            self.stdout.write("Deleted %d objects." % count)
+            with connection.cursor() as cursor:
+                cursor.execute("TRUNCATE TABLE auditlog_logentry;")
```
hramezani, line 31: I think we don't have the `truncate` command in `SQLite`

MHM5000:
maybe a check on DB type before running truncate?

hramezani:
That would be an option. but please consider the ORM will also follow foreign key references and delete objects in other tables.

So, I would suggest implementing this as a flag in `auditlogflush` management command. the default behavior would be still `delete` and by providing the flag, the management command will use the `truncate` instead.

Please also consider adding:

- Tests
- Document for the new flag in https://github.com/jazzband/django-auditlog/blob/master/docs/source/usage.rst#management-commands
- Changelog in https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md

Feel free to open the PR when you are ready.

PR #338, opened 2022-01-04
build(python): remove python 3.6 support

Linkid:
Python 3.6 reaches end of life in December 2021.

This PR fixes CI too.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,5 @@
 # Docs requirements
+django>=3.2,<3.3
 sphinx
 sphinx_rtd_theme
 psycopg2-binary
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    {py37,py38,py39}-django-{22,31}
+    {py37,py38,py39}-django{22,31}
     {py37,py38,py39,py10}-django32
     py38-docs
     py38-qa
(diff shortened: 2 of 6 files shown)


hramezani:
Thanks @Linkid for the patch.
As Django 2.2 will be supported until April 2022, I would suggest keeping Python 3.6 and remove it with Django 2.2

Linkid:
@hramezani: I see what you mean, but Python 3.6 is not supported anymore since December 2021. And even if Django 2.2 supports it, it is like Python 3.5 which is not supported anymore but Django 2.2 supports it officially.

tox.ini
```diff
@@ -29,11 +29,12 @@ basepython =
     py39: python3.9
     py38: python3.8
     py37: python3.7
-    py36: python3.6
 
 [testenv:py38-docs]
 changedir = docs/source
-deps = -rdocs/requirements.txt
+deps =
+    Django>=3.2,<3.3
```
hramezani, line 36: Why you didn't add this to the `docs/requirements.txt`?

Linkid: Seems like all the auditlog package is needed (and installed with tox by default).

docs/requirements.txt
```diff
@@ -1,4 +1,6 @@
 # Docs requirements
+.
```
hramezani, line 2: I think we don't need this line.

hramezani (approved):

Linkid:
Thanks :)

Issue #334, opened 2021-12-14
RemovedInDjango40Warning: ugettext deprecated in django 3 and removed in django 4


wearebasti:
From first tests with Django 4.0 we seem to be "blocked" to upgrade Django versions by django-auditlog, as the pre-release version still uses at least in `auditlog/models.py` an import of `ugettext_lazy` which was deprecated in Django 3 and removed in Django 4.

This should be a pretty simple fix to "just' replace `ugettext_lazy` with `gettext_lazy`, hoping now other issues come up.

Versions:
- django 4.0
- django-auditlog 1.0a1.

Linkid:
Done for the auditlog codebase. Next step, the `django-jsonfield` dependency.

Linkid:
The PR #370 removes the dependency with the warning. So, I can close this issue.

PR #339, opened 2022-01-06
build: replace django-jsonfield with django-jsonfield-backport

Linkid:
The `django-jsonfield` module is not maintained anymore and raises some errors with Django 4.0. So, as a recommendation in this package, and as Django 2.2 is still maintained, `django-jsonfield-backport` will do the job for the `JSONField` field.

Ref #43
Ref #334

PS: this PR is another try to solve what #271 is solving, but by keeping the `JSONField` field.
PPS: `django-jsonfield-backport` will be removed when Django 2.2 will be end of life (Django supports `JSONField` since 3.1).


hramezani:
@Linkid Thanks for this patch. I will take a look later!

hramezani (approved):
LGTM!

PR #335, opened 2021-12-14
Replace ugettext with gettext for django4 compatibility

wearebasti:
This should be a "in-place-replacement" with hopefully no side effects, would be great to have it in the pre-release package, for example as 1.0a2 :) 

This should fix #334

PR #340, opened 2022-01-06
replace ugettext with gettext for django4

hramezani:
Continue to https://github.com/jazzband/django-auditlog/pull/335
I've just rebased the patch.


Linkid:
Hmm… why didn't you just merge the previous PR ? A rebase were only needed if the code were not up to date (and maybe if you were not sure about tests, but it was ok). Furthermore, I think you can edit other's PR if the right option is ticked (ticked by default) ;). So, no need to open a new one.

Linkid (approved):

hramezani:
The merge button was not enabled.
I tried to push to previous PR, i wasn't possible because of permissions. That's why I created a new pr

PR #197, opened 2018-12-11
Fix DeprecationWarning / invalid escape sequence on Python 3.7.x

jerrcs:
Fixes this error whenever running Django commands:

```
/usr/local/lib/python3.7/site-packages/auditlog/models.py:358: DeprecationWarning: invalid escape sequence \.
  add_introspection_rules([], ["^auditlog\.models\.AuditlogHistoryField"])
```

PR #341, opened 2022-01-07
Update django

hramezani:


auditlog/models.py
```diff
@@ -418,7 +418,7 @@ def bulk_related_objects(self, objs, using=DEFAULT_DB_ALIAS):
 try:
     from south.modelsinspector import add_introspection_rules
 
-    add_introspection_rules([], ["^auditlog\.models\.AuditlogHistoryField"])
+    add_introspection_rules([], [r"^auditlog\.models\.AuditlogHistoryField"])
```
Linkid, line 421: Already done in PR #197.

setup.py
```diff
@@ -32,7 +32,6 @@
         "Programming Language :: Python :: 3.10",
         "Framework :: Django",
         "Framework :: Django :: 2.2",
-        "Framework :: Django :: 3.1",
         "Framework :: Django :: 3.2",
```
Linkid, line 35: You can add the django 4.0 classifier ;).

tox.ini
```diff
@@ -1,7 +1,8 @@
 [tox]
 envlist =
-    {py37,py38,py39}-django{22,31}
+    {py37,py38,py39}-django22
     {py37,py38,py39,py10}-django32
+    {py38,py39,py10}-djangomain
```
Linkid, line 5: I think it should be better to explicit the version (`django40`) to know what it is going on without checking another repo.

tox.ini
```diff
@@ -11,8 +12,8 @@ commands =
     coverage xml
 deps =
     django22: Django>=2.2,<2.3
-    django31: Django>=3.1,<3.2
     django32: Django>=3.2,<3.3
+    djangomain: https://github.com/django/django/archive/main.tar.gz
```
Linkid, line 16: Idem: replace `djangomain` with `django40` to be more explicit.

Linkid (changes requested):
Great !

auditlog/models.py
hramezani: I think we can close that one.

setup.py
hramezani: Added!

tox.ini
hramezani: I added Django4.0 to the test pipeline.

hramezani:
Thanks @Linkid for the review. I've addressed your comments!

auditlog/models.py
Linkid: Hmm… I think it is better to merge other's PR instead of redoing it. Check #342 for a quick rebase.

hramezani: you exactly did what I did here :)
The change is already part of this PR. why do we need to have another PR for it?

Linkid: To give credit to the one who did the fix. But if you think it is better to include it in your PR, let's go.

Linkid (approved):

PR #346, opened 2022-01-11
Fix black error.

hramezani:
After https://github.com/jazzband/django-auditlog/commit/ffd8bb1e85fe7ed80d85062e4d5932f28065b84c we have [black error](https://github.com/jazzband/django-auditlog/runs/4773356797?check_suite_focus=true)


Linkid (approved):

Linkid:
Nice catch !

PR #347, opened 2022-01-19
docs(conf): replace pkg_resources with importlib

Linkid:
The doc is build with python 3.8, which includes the `importlib` API to get the package version. Its use is recommanded by `setuptools-scm` for the doc.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,7 @@
+[build-system]
+requires = ["setuptools>=45", "wheel", "setuptools_scm>=6.2"]
+build-backend = "setuptools.build_meta"
+
 [tool.black]
 target-version = ["py37"]
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -9,7 +9,7 @@
 import os
 import sys
 from datetime import date
-from pkg_resources import get_distribution
+from importlib.metadata import version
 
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -1,3 +1,7 @@
+[build-system]
```
hramezani, line 1: Is it necessary to add `build-system`?

Linkid: This is the classical way to package a python module, and documented in the `setuptools_scm` README: https://github.com/pypa/setuptools_scm#pyprojecttoml-usage.

hramezani: If I understand it correctly, your patch replaces `pkg_resources` and uses `importlib` to fetch the version.
So, the question is, do we need this section in `pyproject.toml` for fetching the version?
Or meybe I get it wrong

Linkid: This section is only about build: it follows python and `setuptools_scm` guidelines, and it makes the build more explicit about tools. I added it because it is in the README of `setuptools_scm`, which we are using to find the version of the package. In fact I could put this change in another commit or another PR if you want.

hramezani: I would put it in a separate commit.

Linkid: Done.

hramezani (approved):

PR #351, opened 2022-01-20
docs(readme): update the release section

Linkid:
Need the project lead validation, to be sure ;).

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,4 +42,3 @@ Releases
 5. As a project lead, once the PR is merged, create and push a tag `vX.Y.Z`: this will trigger the release build and a notification will be sent from Jazzband of the availability of two packages (tgz and wheel)
 6. Test the install
 7. Publish the release to PyPI
-9. Update the CHANGELOG with the future version
--- a/README.md
+++ b/README.md
@@ -35,8 +35,12 @@ If you have great ideas for Auditlog, or if you like to improve something, feel 
 Releases
 --------
 
-1. Make sure all tests on `master` are green.
-2. Create a new branch `vX.Y.Z` from master for that specific release.
-3. Bump versions in `setup.py` and `docs/source/conf.py` (docs have 2 places where the versions need to be changed!)
-4. Pull request `vX.Y.Z` -> `master`. 
-5. Pull request `master` -> `stable`. This merge triggers the deploy to pypi.
+1. Make sure all tests on `master` are green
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,8 +35,12 @@ If you have great ideas for Auditlog, or if you like to improve something, feel
...
-2. Create a new branch `vX.Y.Z` from master for that specific release.
-3. Bump versions in `setup.py` and `docs/source/conf.py` (docs have 2 places where the versions need to be changed!)
-4. Pull request `vX.Y.Z` -> `master`. 
-5. Pull request `master` -> `stable`. This merge triggers the deploy to pypi.
+1. Make sure all tests on `master` are green
+2. Create a new branch `vX.Y.Z` from master for that specific release
+3. Update the CHANGELOG release date
+4. Pull request `vX.Y.Z` -> `master`
+5. As a project lead, once the PR is merged, create and push a tag `vX.Y.Z`: this will trigger the release build and a notification will be sent from Jazzband of the availability of two packages (tgz and wheel)
+6. Test the install
```
hramezani, line 43: What does this step mean?

README.md
```diff
@@ -35,8 +35,12 @@ If you have great ideas for Auditlog, or if you like to improve something, feel
...
+7. Publish the release to PyPI
+8. Pull request `master` -> `stable`
```
hramezani, line 45: Do we really need to do this?

README.md
Linkid: Maybe I should be more explicit, yes ^^. One should install the generated wheel and maybe run tests to be sure everything is working as expected.

README.md
Linkid: It is always a good thing to keep a branch with the latest major / minor version to apply patches. Maybe the name is not adapted to our needs. What do you think about that process ? It is the same as Django.

hramezani: I think we don't need this in this project. because we only maintain one version of the package.

hramezani, line 45:
```suggestion
```

README.md
```diff
@@ -35,8 +35,12 @@ If you have great ideas for Auditlog, or if you like to improve something, feel
...
+9. Update the CHANGELOG with the future version
```
hramezani, line 46:
```suggestion
```

hramezani (approved):

Issue #354, opened 2022-02-08
model_instance_diff and smart_str are not realiable with models.JSONField


barrachri:
Hi there!

Long story short:

```python
from django.utils.encoding import smart_str
a = {'description': 'Method form.', 'quantity': '1814.4348944665', 'unit_of_measure': 'bytes', 'unit_price': '144.9156', 'discount_rate': '42.8375', 'tax_rate': '17.06'}
b = {'quantity': '1814.4348944665', 'tax_rate': '17.06', 'unit_price': '144.9156', 'description': 'Method form.', 'discount_rate': '42.8375', 'unit_of_measure': 'bytes'}

# True
assert a == b

# False
assert smart_str(a) == smart_str(b)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError
```

I discovered this by moving from sqlite to postgreSQL, as the order of the data is not guaranteed (I guess because sqlite stores json as text while postgreSQL doesn't).

https://github.com/jazzband/django-auditlog/blob/665217d32f5d07a774fe53cafde6aef36bb4bff5/auditlog/diff.py#L72

https://github.com/jazzband/django-auditlog/blob/665217d32f5d07a774fe53cafde6aef36bb4bff5/auditlog/diff.py#L148

PR #355, opened 2022-02-08
Skip stringify if json

barrachri:
Fixes https://github.com/jazzband/django-auditlog/issues/354

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changes
 
-## 1.0.0 (unreleased)
+## 1.0.0 (2022-01-24)
 
-### Final (unreleased)
+### Final
 
 #### Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 2.0.1
 
 #### Fixes
-- Fix inconsistent change with JSONField ([#355](https://github.com/jazzband/django-auditlog/pull/355))
+- Fix inconsistent changes with JSONField ([#355](https://github.com/jazzband/django-auditlog/pull/355))
 
 ## 2.0.0 (2022-05-09)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changes
 
+## 2.0.1
+
+#### Fixes
+- Fix inconsistent change with JSONField ([#355](https://github.com/jazzband/django-auditlog/pull/355))
+
 ## 2.0.0 (2022-05-09)
 
 #### Improvements
--- a/auditlog/diff.py
+++ b/auditlog/diff.py
@@ -58,25 +58,19 @@ def get_field_value(obj, field):
     :return: The value of the field as a string.
     :rtype: str
     """
-    if isinstance(field, DateTimeField):
-        # DateTimeFields are timezone-aware, so we need to convert the field
-        # to its naive form before we can accurately compare them for changes.
-        try:
+    try:
+        if isinstance(field, DateTimeField):
+            # DateTimeFields are timezone-aware, so we need to convert the field
+            # to its naive form before we can accurately compare them for changes.
             value = field.to_python(getattr(obj, field.name, None))
             if value is not None and settings.USE_TZ and not timezone.is_naive(value):
                 value = timezone.make_naive(value, timezone=timezone.utc)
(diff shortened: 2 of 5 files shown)


auditlog/diff.py
```diff
@@ -67,6 +67,9 @@ def get_field_value(obj, field):
                 value = timezone.make_naive(value, timezone=timezone.utc)
         except ObjectDoesNotExist:
             value = field.default if field.default is not NOT_PROVIDED else None
+    elif isinstance(field, JSONField):
+        try:
+            value = field.to_python(getattr(obj, field.name, None))
```
Linkid, line 72: Exception should be handled

auditlog/diff.py
```diff
@@ -1,6 +1,6 @@
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
-from django.db.models import NOT_PROVIDED, DateTimeField, Model
+from django.db.models import NOT_PROVIDED, DateTimeField, JSONField, Model
```
Linkid, line 3: You should prefer `from django_jsonfield_backport.models import JSONField` instead, for Django compat'.

Linkid (changes requested):
Thanks for your contribution. I let you some comments.

auditlog/diff.py
barrachri: Hi @Linkid, updated.

auditlog/diff.py
barrachri: Also updated this one.

barrachri:
Hi @Linkid, can you give the PR another check? Thanks!

Linkid (changes requested):
Hi! Sorry about that, but with the PR #370, `django-jsonfield-backport` is removed. Then, you can use this line instead:
`from django.db.models import JSONField`

alexsavio:
Sounds good to me. Thanks for the review @Linkid. When are you planning to make a release, though? Thanks a lot!

auditlog/diff.py
```diff
@@ -3,6 +3,7 @@
 from django.db.models import NOT_PROVIDED, DateTimeField, Model
 from django.utils import timezone
 from django.utils.encoding import smart_str
+from django_jsonfield_backport.models import JSONField
```
barrachri, line 6:
```suggestion
from django.db.models import JSONField
```

barrachri:
Thanks @hramezani. Do you think the added tests are enough?

hramezani:
Thanks @barrachri for the tests. But I can see your tests are passing without the fix.
You need to add a test that shows the problem which your patch fixes it.

Also, Please add a changelog entry.

barrachri:
Hi @hramezani you can try to run it again, while disabling the `elif isinstance(field, JSONField):` if branch.

I also added the changelog entry.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changes
 
+## 2.0.1
+
```
hramezani, line 4: let's remove the next version number. we will add it in the next release.
```suggestion
```

auditlog/diff.py
```diff
@@ -67,6 +67,11 @@ def get_field_value(obj, field):
                 value = timezone.make_naive(value, timezone=timezone.utc)
         except ObjectDoesNotExist:
             value = field.default if field.default is not NOT_PROVIDED else None
+    elif isinstance(field, JSONField):
+        try:
+            value = field.to_python(getattr(obj, field.name, None))
+        except ObjectDoesNotExist:
```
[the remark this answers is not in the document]
barrachri: There's no test, _in general_, that covers the `except ObjectDoesNotExist:` part of get_field_value.

I am happy to add one though not entirely sure how to replicate that directly with the model (I believe we would need a migration to replicate it with save()).

But I can call model_instance_diff directly, with 2 instances and one has the attributed deleted.

hramezani (dismissed):
Great work @barrachri :+1:

auditlog/diff.py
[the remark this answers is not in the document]
hramezani: BTW, it is a blocker because coverage fails and I can't merge the PR

barrachri:
Thanks @hramezani, I added a new test and refactor the ObjectDoesNotExist try/except.

hramezani (approved):

PR #364, opened 2022-03-11
Add an index to the timestamp column

gwax:
Many queries, including the default ordering, for the LogEntry
model rely on the timestamp field. Adding an index will ensure
reasonable scalability for large audit logs.


CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
 # Changes
 
 #### Improvements
+
+- index added to the timestamp column
```
Linkid, line 5: This introduced a new schema for the database. I think it should be specified.

gwax: Can you clarify what you mean when you ask for the new schema to be specified?

To my understanding, that is what I am saying here: I have added an index to the timestamp column.

Linkid: Maybe a more explicit sentence could be great, something like: "feat(models): index added to `LogEntry.timestamp`".
I like to see which part of the library is modified when I read changelogs.

gwax: gotcha, how about now?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 #### Improvements
 
+- feat: add db_index to the `LogEntry.timestamp` column
```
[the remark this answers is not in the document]
gwax: done

hramezani (approved):

Linkid (approved):

hramezani:
@alieh-rymasheuski Actually, it is because of https://github.com/jazzband/django-auditlog/commit/128555fa29d71733cbba1201dd399f506e2ebf50

I've prepared a PR to fix the warning.

PR #381, opened 2022-05-31
Fix n+1 query problem

2ykwang:
fix n+1 query problem

n+1
스크린샷 2022-05-31 오후 4 44 09

fixed
스크린샷 2022-05-31 오후 4 44 12

### DIFF
--- a/auditlog/admin.py
+++ b/auditlog/admin.py
@@ -26,5 +26,9 @@ class LogEntryAdmin(admin.ModelAdmin, LogEntryAdminMixin):
         # As audit admin doesn't allow log creation from admin
         return False
 
+    def get_queryset(self, request):
+        queryset = super().get_queryset(request).select_related("content_type", "actor")
+        return queryset
+
 
 admin.site.register(LogEntry, LogEntryAdmin)


hramezani:
Thanks @2ykwang :+1: 

Please add a changelog entry for this change.

2ykwang:
@hramezani thanks

done

hramezani (approved):

auditlog/admin.py
```diff
@@ -26,5 +26,9 @@ def has_add_permission(self, request):
         # As audit admin doesn't allow log creation from admin
         return False
 
+    def get_queryset(self, request):
+        queryset = super().get_queryset(request).select_related("content_type", "actor")
```
alieh-rymasheuski, line 30: @2ykwang, why not use [`list_select_related`](https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_select_related)?

hramezani: Yes, @2ykwang Could you please give this a try and compare the numbers?

PR #382, opened 2022-05-31
Auditlog admin use list_select_related

2ykwang:
improve

#381


alieh-rymasheuski:
Merging even though `codecov/project` failed - it's a normal phenomenon when the code gets removed or shortened.

PR #392, opened 2022-06-28
Fix & improve running tests with tox

alieh-rymasheuski:
There are three unrelated fixes and improvements in this pull request:
1. A typo in tox env list is fixed.
2. An expected warning is silenced in the tests.
3. Test configuration is adjusted to allow running `tox -p`.

The third item improves the total duration of tests (as measured by `time`) on the local machine drastically:

| command | sequential (no `-p` flag) | parallel (with `-p` flag) |
|----------|---------------------------|---------------------------|
| `tox` | 2:53.37 | 59.92 |
| `tox -r` | 4:55.11 | 1:37.75 |

Note: `tox -p` runs 8 envs in parallel on my machine by default. This number and thus the timings may be different depending on the configuration of your machine.

### DIFF
--- a/auditlog_tests/test_settings.py
+++ b/auditlog_tests/test_settings.py
@@ -29,7 +29,9 @@ MIDDLEWARE = (
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.postgresql",
-        "NAME": os.getenv("TEST_DB_NAME", "auditlog_tests_db"),
+        "NAME": os.getenv(
+            "TEST_DB_NAME", "auditlog" + os.environ.get("TOX_PARALLEL_ENV", "")
+        ),
         "USER": os.getenv("TEST_DB_USER", "postgres"),
         "PASSWORD": os.getenv("TEST_DB_PASS", ""),
         "HOST": os.getenv("TEST_DB_HOST", "127.0.0.1"),
--- a/auditlog_tests/tests.py
+++ b/auditlog_tests/tests.py
@@ -1,6 +1,7 @@
 import datetime
 import itertools
 import json
+import warnings
 from unittest import mock
 
 from dateutil.tz import gettz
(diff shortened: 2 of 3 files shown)


auditlog_tests/test_settings.py
```diff
@@ -29,7 +29,9 @@
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.postgresql",
-        "NAME": os.getenv("TEST_DB_NAME", "auditlog_tests_db"),
+        "NAME": os.getenv(
+            "TEST_DB_NAME", "auditlog" + os.environ.get("TOX_PARALLEL_ENV", "")
```
hramezani, line 33: Why do we need to change the database name?
it was `auditlog_tests_db` before if we don't provide the `TEST_DB_NAME`. and we already defined valued for `TEST_DB_NAME` for [Github ci](https://github.com/jazzband/django-auditlog/blob/4f4699bf6b48a02ee288f0ba28d63b8d7e851bac/.github/workflows/test.yml#L63) and [tox](https://github.com/jazzband/django-auditlog/blob/4f4699bf6b48a02ee288f0ba28d63b8d7e851bac/tox.ini#L25)

auditlog_tests/tests.py
```diff
@@ -600,6 +601,18 @@ class DateTimeFieldModelTest(TestCase):
...
+        self._context = warnings.catch_warnings()
+        self._context.__enter__()
+        warnings.filterwarnings(
+            "ignore", message=".*naive datetime", category=RuntimeWarning
+        )
+
+    def tearDown(self):
+        self._context.__exit__()
+        super().tearDown()
+
```
hramezani, line 615: is it possible to just ignore the warning for that specific test that created the warning?

tox.ini
```diff
@@ -1,11 +1,13 @@
...
-    {py37,py38,py39,py10}-django32
-    {py38,py39,py10}-django{40,main}
+    {py37,py38,py39,py310}-django32
+    {py38,py39,py310}-django{40,main}
     py37-docs
     py38-lint
 
 [testenv]
+setenv =
+  COVERAGE_FILE={toxworkdir}/.coverage.{envname}
```
hramezani, line 10: Could you please elaborate on this line? why do we need it?

auditlog_tests/test_settings.py
alieh-rymasheuski: When tox runs tests in parallel, it passes `TOX_PARALLEL_ENV` environment variable to every parallel worker. We use this value to make sure that every parallel worker uses its own database and can therefore drop or recreate it and run the migrations on it without causing race conditions with other workers.

auditlog_tests/tests.py
alieh-rymasheuski: This warning is caused by multiple tests in this test class. I decided to effectively wrap every method in this class in `with catch_warnings`.

hramezani (approved):
Thanks @alieh-rymasheuski :+1:

Issue #410, opened 2022-08-05
Need to be able to access serialized data from changed object


sum-rock:
We need to be able to access a serialized dictionary of an object's field values following a change. I'd like to add a `serialized_data` JSON field to the `LogEntry` model. That field will be populated by the receivers which will implement `django.core.serializers` to serialize the new state of the object after a change.

Having a historical record of object states will help us to be able to recover from several types of potential data loss events and aid us in debugging system artifacts. It is also critical to help us to determine calculated values for post hoc analysis and reporting. For example, if we need to determine who had _x_ insurance and lived in _y_ state at _z_ time, we can use the audit log to determine what the person had listed as their state of residence at time _z_ without needing a separate one to many historical model for state of residence. 

There are a lot of helpful public facing APIs that could be added to the project using this field as a foundation (e.g., `get_field_value_at_time(field_name, timestamp)`) but the first step is to create the `serialized_data` field itself. Also note that I am aware there are other packages out there, such as [django-reversions](https://github.com/etianen/django-reversion), that do this. However, in our experience they don't fit our needs as a first-class audit log.

I am adding this functionality to our fork but would love it if you would consider adding it to the project. Please let me know if there is anything I can do to make that more likely. Look for my PR later today!

cc @hramezani

rposborne:
And to pile on, this is not an attempt to "restore" the data for a given point in time.  It is only to capture the complete state instead of only a text representation of a given object as currently represented by the `obj_repr` field.

hramezani:
@sum-rock Thanks for creating this feature request.
I think I got the overall idea of the feature. adding a `serialized_data` JSON field in the `LogEntry` model and saving the object after change in a json format into the field.
Then you want to use the new field for finding the value of some field.

> For example, if we need to determine who had x insurance and lived in y state at z time, we can use the audit log to determine what the person had listed as their state of residence at time z without needing a separate one to many historical model for state of residence.

Could you please provide a small example here? I mean by having a model and providing the data for the new `serialized_data` field and explaining how you do the query.

I am not sure about the feature. I would like to heat @Linkid and @alieh-rymasheuski opinions here.

Note: We can have the feature in disable mode by default and make it enable in settings.

sum-rock:
@hramezani I figured the best way to show you my thoughts was in the PR. Take a look at what I've submitted and let me know what you all think. I'm happy to make changes if it has a better chance of getting merged!

rposborne:
[...]

PR #412, opened 2022-08-08
Add serialized object field

sum-rock:
This work creates the basic implementation of the `LogEntry.serialized_object` field. This field tracks the state of an object following a create, update, or delete action. A manager method is also added to retrieve the value of an object's field at a given date.

Efforts were taken to make sure this addition is not disruptive to current installations of django-auditlog. Specifically, this is an opt-in feature per model. Users are able to opt in by adding the `serialize_data=True` kwarg to the model registration.

This relates to issue [410](https://github.com/jazzband/django-auditlog/issues/410).


sum-rock:
This PR contains a migration. If the wonderful maintainers of django-auditlog would help my team out tremendously by adding this to their project, we would need to be mindful of any other PR's with migrations that might also make it into the release. For example, [#407](https://github.com/jazzband/django-auditlog/pull/407) also contains a migration and we'd need to coordinate that with this one so that they don't have shared dependencies.

I'm also aware that I'd need to add something to the change log here. Let me know what you all think of this PR and if you're feeling positive about it, I'll write something up.

hramezani:
Thanks @sum-rock for the PR :+1: 

I am still not sure about the idea of the feature.
I will back to the PR later. Unfortunately, I don't have time to check it now.

@alieh-rymasheuski @Linkid It would be great if you have time to take a look at this PR.

auditlog/models.py
```diff
@@ -6,15 +6,18 @@
...
 from django.db.models import Q, QuerySet
 from django.utils import formats, timezone
 from django.utils.encoding import smart_str
 from django.utils.translation import gettext_lazy as _
 
+from .history import FieldHistoryValueManagerMixin
 
-class LogEntryManager(models.Manager):
+
+class LogEntryManager(FieldHistoryValueManagerMixin, models.Manager):
```
sum-rock, line 20: I took the approach of a Mixin here so that I could move the "historical state" references to a new file. My thought is that if you decide to not include the manager features, it will be easier for me to maintain the fork by having the distinct code in a separate file (or at least as much of it is possible.)

auditlog/models.py
```diff
@@ -6,15 +6,18 @@
...
 from django.db.models import Q, QuerySet
 from django.utils import formats, timezone
 from django.utils.encoding import smart_str
 from django.utils.translation import gettext_lazy as _
 
+from .history import HistoricalStateLookupManagerMixin
 
-class LogEntryManager(models.Manager):
+
+class LogEntryManager(HistoricalStateLookupManagerMixin, models.Manager):
```
sum-rock, line 20: The idea behind putting a mixin here is so that if you don't want to include the historical state lookups in the manager, It will be easier for me to maintain a separate fork with those features included if the distinct code is in a different file.

sum-rock:
Given some of the feedback here, I've made a few adjustments. I'm still optimistic that you all may still be favorable to including the `serialized_data` field in the auditlog project. When I think about the typical use cases of an audit log, I think that having an option to access a full account of an objects state, either prior to or following a change, would be helpful in understanding not only what changes were made, but also why they might have been made. This is a concept that is embraced by other audit logging systems such as [Papertrail](https://github.com/paper-trail-gem/paper_trail). However, there isn't a good option for Django that we are aware of. Django auditlog is so darn close to having everything that anyone would ever need in an audit log! We just want to help put the cherry on top of the Sunday here.

I can understand if the contents of the `HistoricalStateLookupManagerMixin` is too heavy handed (although I think people would use it to avoid boilerplate code and more easily find the object or field states they are looking for). So if that is not in keeping with the vision of django auditlog, I can definitely extract it from this PR.

Please let me know if you all have any other concerns that could be mitigated.

hramezani:
Thanks @sum-rock for the update. I will take a look whenever I have time.

auditlog/history.py
```diff
@@ -0,0 +1,131 @@
...
+        from .models import LogEntry
+
+        return LogEntry.objects.get(id=self.log_entry_id)
+
+
+class HistoricalStateLookupManagerMixin:
+    def get_for_object(self, instance):
+        raise NotImplementedError(f"{self.__class__} requires a get_for_object method.")
+
+    def get_object_state_at_timestamp(self, instance, timestamp):
```
hramezani, line 50: Why we don't return the found log entry in this function instead of `HistoricalObjectState`?

What would be the benefit of `HistoricalObjectState` and `HistoricalObjectState`?

auditlog/history.py
```diff
@@ -0,0 +1,131 @@
...
+
+        # Return a historical object state
+        return HistoricalObjectState(
+            log_found=True,
+            serialized_fields=serialized_fields,
+            timestamp=log.timestamp,
+            log_entry_id=log.id,
+        )
+
+    def get_field_state_at_timestamp(self, instance, field_name, timestamp):
```
hramezani, line 88: Seems this method here uses the output of `get_object_state_at_timestamp` and looks for the value of `field_name`? right?

hramezani:
To me, the change has two parts

1. Adding a JSON field to store the JSON repr of the object. (sounds good)
2. The `HistoricalStateLookupManagerMixin` class, is about how we use the stored data. there are two methods there for finding the log record and finding a field value from the log at a specific time.

I think we can remove the `HistoricalStateLookupManagerMixin` from this path because:

- I think the idea of this library is to store the changes, not how to use them.
- Finding a log entry at a given timestamp is not a big thing and users of the library can easily do it
- The currently added functions are not complete. we only have a filter for one timestamp with `__lte` and maybe we need another filter for `__gte`, `__lt`, `__gt`, ... . we may want to get multiple fields value not just one field value. we may want to apply filters based on user and action as well. In the end I would say the functions are not generic enough to cover all the needs

I would like to know what do you think @sum-rock and @rposborne?

auditlog/history.py
sum-rock: Yup. I figured we could break that into two pieces in case people want to use the object state instead of the field state. The real purpose here is to provide an in built mechanism to retrieve the field or object state and have a flag to show if the field or object was found in the logs. It's important to put these checks in the returned results to distinguish something that was found with a result of `None` and something that is `None` because it wasn't present in `LogEntry`

hramezani:
Thanks for your understanding @sum-rock

hramezani:
Why did you add the changes from the other PR here?
It makes it hard to review. We can first review this one and it can be merged to master because it is backward compatible. Then you could refresh the other PR to fix the migration dependency

sum-rock:
@hramezani I really apologize for that. I had some git madness on my end and unintentionally made that happen. It's fixed now. I was hoping it would be fixed before you'd notice 🙈 .

hramezani:
Please add documentation into [usage](https://github.com/jazzband/django-auditlog/blob/master/docs/source/usage.rst) after `Many-to-many fields` section.

auditlog/registry.py
```diff
@@ -78,7 +84,8 @@ def register(
         :param mapping_fields: Mapping from field names to strings in diff.
         :param mask_fields: The fields to mask for sensitive info.
         :param m2m_fields: The fields to handle as many to many.
-
+        :param serialize_data: Option to include a dictionary of the objects state in the auditlog.
+        :param serialize_kwargs: Optional kwargs to pass to django serializer
```
hramezani, line 88:
```suggestion
        :param serialize_kwargs: Optional kwargs to pass to Django serializer
```

auditlog/registry.py
```diff
@@ -68,6 +72,8 @@ def register(
         mapping_fields: Optional[Dict[str, str]] = None,
         mask_fields: Optional[List[str]] = None,
         m2m_fields: Optional[Collection[str]] = None,
+        serialize_data: bool = False,
+        serialize_kwargs: Optional[dict] = None,
```
hramezani, line 76: cant we set the default value to `{}`?

```suggestion
        serialize_kwargs: Dict[str, Any] = {},
```

auditlog/registry.py
```diff
@@ -91,6 +98,14 @@ def register(
             mask_fields = []
         if m2m_fields is None:
             m2m_fields = set()
+        if serialize_kwargs is None:
+            serialize_kwargs = {}
```
hramezani, line 102: Maybe we don't need this block here if we set the default value to `{}`

auditlog_tests/tests.py
```diff
@@ -1104,6 +1107,13 @@ def test_register_from_settings_register_models(self):
         self.assertEqual(fields["include_fields"], ["label"])
         self.assertEqual(fields["exclude_fields"], ["text"])
 
+    def test_registration_error_if_bad_serialize_params(self):
+        with self.assertRaises(AuditLogRegistrationError):
```
hramezani, line 1111: Please test the error message here as well. like https://github.com/jazzband/django-auditlog/blob/a24b79af0c558c50381e752377e3003f0c53b564/auditlog_tests/tests.py#L1027-L1030

auditlog/registry.py
alieh-rymasheuski: No, this will store the reference to the mutable dict in the function object! Check this:
```python
def add(k, v, d={}):
    d[k] = v
    return d

print(add("key1", "value1"))  # -> {"key1": "value1"}
print(add("key2", "value2"))  # -> {"key1": "value1", "key2": "value2"}
```

auditlog/models.py
```diff
@@ -208,6 +212,18 @@ def _get_pk_value(self, instance):
...
+    def _get_serialized_data_or_none(self, instance):
+        from .registry import auditlog
+
+        registry_opts = auditlog.get_model_fields(instance.__class__)
+        if not registry_opts["serialize_data"]:
+            return None
+
+        kwargs = registry_opts.get("serialize_kwargs", {})
+        kwargs.setdefault("fields", [x.name for x in instance._meta.fields])
+        data = json.loads(serializers.serialize("json", (instance,), **kwargs))
```
alieh-rymasheuski, line 224: We have the logic which masks the values of certain fields (see `mask_fields` setting), we should respect this setting in all our serialization because they may be used for privacy or legal concerns.

auditlog/models.py
```diff
@@ -208,6 +212,18 @@ def _get_pk_value(self, instance):
...
+        return next(iter(data), None)
```
alieh-rymasheuski, line 225: What is the case in which `data` can be empty?

auditlog/registry.py
hramezani: Ah yes! forgot about this. Thanks for pointing.

sum-rock: How about that. Very interesting. The more you know the more you know you don't know anything.

auditlog/history.py
sum-rock: removed.

auditlog/history.py
```diff
@@ -0,0 +1,131 @@
...
+        return LogEntry.objects.get(id=self.log_entry_id)
+
+
+class HistoricalStateLookupManagerMixin:
+    def get_for_object(self, instance):
+        raise NotImplementedError(f"{self.__class__} requires a get_for_object method.")
+
+    def get_object_state_at_timestamp(self, instance, timestamp):
+        """
+        Get state of an instance at a given timestamp
```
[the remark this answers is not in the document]
sum-rock: removed.

auditlog/history.py
sum-rock: removed

auditlog/registry.py
sum-rock: 👍  fixed.

auditlog_tests/tests.py
sum-rock: Good call. fixed.

auditlog/registry.py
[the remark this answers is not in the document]
sum-rock: Love me some good and specific type hints. Fixed.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changes
 
+#### Improvements
+
+- feat: Add `serialized_data` field on `LogEntry` model. ([412](https://github.com/jazzband/django-auditlog/pull/412))
```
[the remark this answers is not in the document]
sum-rock: Oops. fixed.

auditlog/models.py
```diff
@@ -208,6 +212,18 @@ def _get_pk_value(self, instance):
             pk = self._get_pk_value(pk)
         return pk
 
+    def _get_serialized_data_or_none(self, instance):
+        from .registry import auditlog
+
+        registry_opts = auditlog.get_model_fields(instance.__class__)
```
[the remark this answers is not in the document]
sum-rock: Good call. I hadn't fully realized the intention of that method until I read this. I've added another method to the `AuditlogModelRegistry` to specifically retrieve the serialization options. How does that look to you?

auditlog/models.py
sum-rock: Well... I may have went overboard on this one but I do like a good challenge. I can appreciate the need to mask values. I currently work in a regulated industry and we may need this feature too. I've added a `MaskedDictionary` object that will recursively search for any flat and nested keys as specified by the `mask_fields` kwarg. Nesting can be specified with key pathing using a double underscore convention. There's a good test case that shows this better than I can explain it in writing.

This might be excessive. Let me know if that is too much for the requirements here. My thought was that if there is any nested serialization for related fields this recursive approach would be necessary. I didn't end up finding a great solution for nested related serialization though. See my points on the relevant comment for that.

auditlog/models.py
sum-rock: Probably never. I've changed the approach here.

auditlog/models.py
```diff
@@ -208,6 +212,18 @@ def _get_pk_value(self, instance):
...
+        if not registry_opts["serialize_data"]:
+            return None
+
+        kwargs = registry_opts.get("serialize_kwargs", {})
+        kwargs.setdefault("fields", [x.name for x in instance._meta.fields])
```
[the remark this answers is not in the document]
sum-rock: Part of the reason my team is searching for a replacement for our current auditing system is because we need to be able to NOT create new audit records when certain field change but we WANT them included in the object serialization. Consider the following example...

We have a model for `Person` and the state of each person instance is examined every _t_ minutes for changes that may require sending the state of the person to an external system. That is, we need to sync each person with an external system every time specific conditions are met and we examine those conditions every _t_ minutes. To track this, we store a value on the `Person` instance for `last_checked_for_sync` and `last_synced`. We do NOT need to bloat our database with audit logs for each examination when the only field that changes is incrementing `last_checked_for_sync` but we do want to have that information if we look back at historical logs and examine real changes. 

I think we have some options here though. Consider that in the `serialize_kwargs` one can already add an argument for `fields` and this will be passed to the native Django serializer. Therefore if one wants to limit the fields being serialized, he/she can specify them there. If that's not intuitive enough there is another option. We could add an argument in `serialize_kwargs` for choosing fields based on the the model field selections. Something like `use_auditlog_fields_only`? On the other hand, I quite like that the current `serialize_kwargs` are all passed directly to the Django serializer though and creating our own would betray that.

If you feel strongly that we need to only be serializing fields based on what the `exclude_fields` and `include_fields` shows, then I would be reluctantly okay with flipping the script and saying that if one really needs to override that behavior then he/she would need to specify all the model fields in the `serialize_kwargs` `fields` argument. It would just be a pain for me and others in my situation.

auditlog/models.py
```diff
@@ -39,6 +40,9 @@ def log_create(self, instance, **kwargs):
             )
             kwargs.setdefault("object_pk", pk)
             kwargs.setdefault("object_repr", smart_str(instance))
+            kwargs.setdefault(
+                "serialized_data", self._get_serialized_data_or_none(instance)
+            )
```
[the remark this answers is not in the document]
sum-rock: FK's are handled natively by Django's core serializer. As for the other relationships...

Object serialization can get a little crazy. I think I'm trying to walk a fine line between storing the state of the object being changed but not attempting to store the state of the system. I recognize that this boundary can be murky when we consider M2M through tables and reverse FK relationships in general. 

I think if we wanted to expand the serialization beyond the scope of the immediate record, then we'd need to include an option to do something like the following. 

1. Examine meta properties of each of the objects fields.
2. For every reverse FK relationship or M2M field found:
    a. Query the related objects
    b. For each object found
        i. Serialize the result
        ii. Assemble a list of serialized dictionaries
    c. Bolt the list of dictionaries onto the field representation.

From my perspective, handling reverse FK relationships is not really worth it. Our team will simply log changes on the related objects separately so we can thread the data together later if necessary. I would be interested though if you thought this juice was really worth the squeeze (did I mention I like challenges). Alternatively, maybe you know about a more elegant approach for implementing nested serialization for reverse FK relationships?

sum-rock:
@hramezani @alieh-rymasheuski Thank you for the code review. I appreciate your feedback and I've made effort to make sure I have addressed each of your comments. There are a few things you may want to take a look at. In particular, the changes around `mask_fields` and my responses to the comments about what fields to include in the serialized data and my thoughts about serializing reverse FK relationships.

Again, I really appreciate the time you both spend in maintaining this project and the diligence you show in your efforts. Enjoy the weekend!

auditlog/models.py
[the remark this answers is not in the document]
hramezani: As the new `serialized_data` is for saving the current state of the object, I would suggest including all fields and adding an option to limit it to auditlog field only. for example `use_auditlog_fields_only`.

Please remember we don't want to make it very complex and the main idea of the library is auditing

What do you think @alieh-rymasheuski

auditlog/models.py
hramezani: Thanks @sum-rock for implementing `MaskedDictionary`. I think it would be better to keep the mask field implementation simple and just mask the field in the model not related models.

auditlog/models.py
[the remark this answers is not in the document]
sum-rock: Sounds good. I can add an option for `use_auditlog_fields_only` to be set in `serialize_kwargs`. I will do my best to keep it simple!

auditlog/models.py
sum-rock: @hramezani I've removed the `MaskedDictionary` and implemented something much simpler. Let me know if this looks acceptable. I also moved this work over to the manager. It seemed to be the right change considering the other adds around ``serialize_auditlog_fields_only``.

auditlog/models.py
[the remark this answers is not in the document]
sum-rock: @hramezani I've added an implementation of ``serialize_auditlog_fields_only``. Hopefully this works for everyone who needs that option. I wanted to keep the ``serialize_kwargs`` reserved for arguments that can be passed directly to the Django core serializer. I think its more discoverable and future proof that way but I'm open to nesting it within the ``serialize_kwargs`` if you feel differently.

docs/source/usage.rst
```diff
@@ -115,6 +115,38 @@ Note that when the user changes multiple many-to-many fields on the same object
 
 .. versionadded:: 2.1.0
 
+**Serialized Data**
+
+The state of an object following a change action may be optionally serialized and persisted in the ``LogEntry.serialized_data`` JSONField.  To enable this feature for a registered model, add ``serialize_data=True`` to the kwargs on the ``auditlog.register(...)`` method. Object serialization will not occur unless this kwarg is set.
```
hramezani, line 120:
```suggestion
The state of an object following a change action may be optionally serialized and persisted in the ``LogEntry.serialized_data`` JSONField. To enable this feature for a registered model, add ``serialize_data=True`` to the kwargs on the ``auditlog.register(...)`` method. Object serialization will not occur unless this kwarg is set.
```

auditlog/models.py
```diff
@@ -208,6 +215,65 @@ def _get_pk_value(self, instance):
...
+        if not opts["serialize_data"]:
+            return None
+
+        kwargs = opts.get("serialize_kwargs", {})
+        kwargs.setdefault("fields", None)
+        data = dict(json.loads(serializers.serialize("json", (instance,), **kwargs))[0])
+        model_fields = auditlog.get_model_fields(instance.__class__)
+
+        if opts["serialize_auditlog_fields_only"]:
+            data = self._filter_serialized_fields_to_eligible(data, model_fields)
```
hramezani, line 231: Isn't it possible to check this option before serializing the data? I mean by adding the fields to `kwarg`? Then we would have one less loop over all data.

docs/source/usage.rst
sum-rock: done.

auditlog/models.py
sum-rock: @hramezani it sure is! Good thinking. Sometimes I can't see the forest from the trees when I'm moving code around. Check it out now.

auditlog/models.py
```diff
@@ -208,6 +215,61 @@ def _get_pk_value(self, instance):
...
+        include_fields = model_fields["include_fields"]
+        exclude_fields = model_fields["exclude_fields"]
+        all_field_names = [field.name for field in instance._meta.fields]
+
+        if not include_fields and not exclude_fields:
+            return all_field_names
+
+        include_fields = include_fields or all_field_names
+        filtered_field_names = set(all_field_names).intersection(include_fields)
+        filtered_field_names = filtered_field_names.difference(exclude_fields)
```
hramezani, line 252: I think we don't need an intersection between `all_field_names` and `include_fields` because it's already happening in line before. am I wrong?
```suggestion
        filtered_field_names = set(filtered_field_names).difference(exclude_fields)
```

sum-rock: Good eye. You are again correct. We can remove a few lines of logic there.

hramezani:
@alieh-rymasheuski as you left some comments on this PR, It would be great if you check the updates here.

auditlog/models.py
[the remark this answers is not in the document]
alieh-rymasheuski: > handling reverse FK relationships is not really worth it

From my perspective, neither is handling M2M fields. In my experience, the fewer relationships are serialized with an object the better. :)

auditlog/models.py
```diff
@@ -1,18 +1,22 @@
...
 from django.contrib.contenttypes.models import ContentType
+from django.core import serializers
 from django.core.exceptions import FieldDoesNotExist
 from django.db import DEFAULT_DB_ALIAS, models
 from django.db.models import Q, QuerySet
 from django.utils import formats, timezone
 from django.utils.encoding import smart_str
 from django.utils.translation import gettext_lazy as _
 
+from .diff import mask_str
```
alieh-rymasheuski, line 18: Please use absolute imports, relative imports are not used in this repository.

auditlog/models.py
```diff
@@ -208,6 +215,58 @@ def _get_pk_value(self, instance):
             pk = self._get_pk_value(pk)
         return pk
 
+    def _get_serialized_data_or_none(self, instance):
```
alieh-rymasheuski, line 218: Can this method still return a `None`? If not, the `none` should be eliminated from the method name.

auditlog/models.py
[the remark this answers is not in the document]
alieh-rymasheuski: Thank you.

auditlog/models.py
```diff
@@ -208,6 +215,58 @@ def _get_pk_value(self, instance):
...
+
+        if opts["serialize_auditlog_fields_only"]:
+            kwargs.setdefault(
+                "fields", self._get_applicable_model_fields(instance, model_fields)
+            )
+
+        data = dict(json.loads(serializers.serialize("json", (instance,), **kwargs))[0])
+
+        if model_fields["mask_fields"]:
+            data = self._mask_serialized_fields(data, model_fields)
```
alieh-rymasheuski, line 236: It's cleaner to pass `model_fields["mask_fields"]` to this method.

auditlog/models.py
```diff
@@ -208,6 +215,58 @@ def _get_pk_value(self, instance):
...
+        include_fields = model_fields["include_fields"]
+        exclude_fields = model_fields["exclude_fields"]
+        all_field_names = [field.name for field in instance._meta.fields]
+
+        if not include_fields and not exclude_fields:
+            return all_field_names
+
+        return list(set(include_fields or all_field_names).difference(exclude_fields))
+
+    def _mask_serialized_fields(
```
alieh-rymasheuski, line 252: What if model A has a foreign key to model B, and model B is configured with `mask_fields`? :)

Can we add a note somewhere in the docs that `mask_fields` doesn't affect fields embedded in serialized related objects? A user may assume that and we should preempt this assumption.

sum-rock: I have a few thoughts here.

1. Forward relationships would simply show the FK unless the developer takes specific actions by setting up natural keys. Using confidential information within natural keys would show up in a whole lot of places and would therefore be unadvisable beyond the scope of the auditlog.
2. Serializing reverse relationships into nested lists are not supported in this PR and therefore `mask_fields` from those related tables will never show up in the serialized data.
3. Most importantly, I had added support for nested masked values from related tables in [this commit](https://github.com/jazzband/django-auditlog/pull/412/commits/23541cf7fef8a7d50c1534f4d8741d11ed691936) but was asked to remove it due to unnecessary complexity (which I tend to agree is the correct decision).

auditlog/models.py
sum-rock: It still returns none.

```python
    opts = auditlog.get_serialize_options(instance.__class__)
    if not opts["serialize_data"]:
        return None
```

auditlog/models.py
sum-rock: Yup yup! Done.

auditlog/models.py
sum-rock: 👍  done.

auditlog/models.py
alieh-rymasheuski: > Forward relationships would simply show the FK

Ah, I misunderstood your other comment and thought that the forward relationship would have had a complete (or incomplete :upside_down_face:) representation of the related object. Please ignore this thread then. :+1:

alieh-rymasheuski:
@hramezani, passing this pull request back to you. I don't have objections to merging this pull request.

hramezani:
> @hramezani, passing this pull request back to you. I don't have objections to merging this pull request.

Thanks @alieh-rymasheuski for the review. I will take a final look later and will merge it.

sum-rock:
@hramezani Before you guys merge. I'm looking into something that's coming up with implementing this in our system. The issue occurs if datetimes are written as strings on in-memory model fields. Without having the serialized_data field, the DB will type cast successfully. However  when updating, the serializer will attempt to serialize a string as if it is a datetime and fail. Please standby as I am currently trouble shooting. I just want more information before you merge.

sum-rock:
@hramezani @alieh-rymasheuski I was running our testing suite with these changes this morning and noticed a few unexpected errors. After fully untangling the stacktrace this afternoon I believe I've found the issue and put together a correction. I do think it is best to put the correction in this PR rather than relying on developers to correct in their own code. Avoiding any disruption, even if users need to opt into this new feature, is in everyone's interest.

Take a look at what I've written and let me know your thoughts. I've tried to include an explanation in the doc string and commit message so I'll avoid repeating it here. One thing I might add is that the try/except around the `deepcopy` was done to avoid raising TypeErrors for missing arguments that are required when attempting to copy some objects. The `instance_copy` is preferred because I'd like to avoid mutating something before saving if necessary but I don't think it's critical because the model save and model instantiation processes will do a similar type transformation in the same way anyway. 

Thanks again for your time and dedication, our entire team is very grateful.

auditlog/models.py
```diff
@@ -208,6 +216,80 @@ def _get_pk_value(self, instance):
...
+        self, data: Dict[str, Any], mask_fields: List[str]
+    ) -> Dict[str, Any]:
+        masked_data = data.copy()
+        all_field_data = dict(masked_data.pop("fields"))
```
hramezani, line 281: Do we really need to convert the `fields` content to dict?

auditlog/models.py
```diff
@@ -208,6 +216,80 @@ def _get_pk_value(self, instance):
...
+
+        if not include_fields and not exclude_fields:
+            return all_field_names
+
+        return list(set(include_fields or all_field_names).difference(exclude_fields))
+
+    def _mask_serialized_fields(
+        self, data: Dict[str, Any], mask_fields: List[str]
+    ) -> Dict[str, Any]:
+        masked_data = data.copy()
```
hramezani, line 280: Do we really need to make a copy of the data? can't we change the data itself? I can see we make another copy in `_get_copy_with_python_typed_fields` as well. as you know the copy is a heavy operation. can't we avoid copying?

auditlog/models.py
```diff
@@ -208,6 +216,80 @@ def _get_pk_value(self, instance):
             pk = self._get_pk_value(pk)
         return pk
 
+    def _get_serialized_data_or_none(self, instance):
+        from .registry import auditlog
```
hramezani, line 220:
```suggestion
        from auditlog.registry import auditlog
```

hramezani:
> @hramezani why milestone 2.2? This seems to be a backwards-compatible change so it can be include in a 1.X feature release.

@alieh-rymasheuski why `1.x`? Latest version for `django-auditlog` is `2.1.1` https://pypi.org/project/django-auditlog/
I would like to include this change in the next release `2.2`

auditlog/models.py
sum-rock: Done.

auditlog/models.py
sum-rock: My thought was that we might want to attempt to set the type and error here rather than sending an unknown type to  the iterator. I suppose it doesn't actually matter where the TypeError would happen. And I'm not aware of any case where the serializer would return a different shape without the nested fields dictionary either. I've removed the conversion.

hramezani (approved):

hramezani:
Thanks @sum-rock for your effort!

PR #420, opened 2022-08-25
Change diff to evaluate PKs for FK relationships

sum-rock:
~~I am going to write up an issue to relate to this PR in the next day or two. At the moment, I'm focused on getting this fix merged into our fork to resolve some bloating on our LogEntry table.~~ (Done, see attached issue)

Take a look and let me know what you think of this fix. I did look through the current issues and couldn't find anyone else complaining about this so maybe we've got some anti-patterns going on in our code. The issue is only happening in some of our more complex systems but I'm sure we're not that unique. I'll highlight the test case that shows the fix.

After letting this sit for an hour or two I realized this change will fix another more insidious bug. Currently, in cases where there are more than one object with the same string representation (e.g., two or more "John Smith" objects in the `Customer` model) and a related object is updated to a different object with the same string representation (e.g., an update to a `PhoneNumber` object "+1555-555-5555" to belong to a different "John Smith" `Customer` object) auditlog is not able to capture that update. This is because because the `diff` method is comparing object strings rather than primary keys and there is no guarantee that the `__str__` method is returning a unique value. Therefore, a portion of updates to `ForeignKey` fields are likely not being captured by auditlog on current installations.

In [this commit](https://github.com/jazzband/django-auditlog/pull/420/commits/c7405fa4c02ca618979e61a2f4d09d44ff0dcc0a) you can see a test case that demonstrates the flaw and how it is resolved by this change.
__ __

The diff method now evaluates the primary keys for changes to determine
if a new LogEntry should be created. Previously, the diff method was
evaluating the string representation of the object. This was flawed
because cases can occur when a parent object has in memory changes to
its string string representation and the child related object is saved
prior to these in memory changes being persisted. In these cases a new
LogEntry object would be created erroneously. This cases is asserted
with a test and a regression test will verify the bug.

The consequence of these updates is that the ``LogEntry.changes`` field
now stores primary keys rather than string representations for related
objects. To keep the changes dictionary display unaffected by this
update, a method was added to the ``LogEntry`` model. This method looks
up the object display string from the stored foreign key. Exceptions
were written to handle backwards compatibility.

[Issue](https://github.com/jazzband/django-auditlog/issues/421)

### DIFF
--- a/auditlog/diff.py
+++ b/auditlog/diff.py
@@ -74,6 +74,10 @@ def get_field_value(obj, field):
                 value = django_timezone.make_naive(value, timezone=timezone.utc)
         elif isinstance(field, JSONField):
             value = field.to_python(getattr(obj, field.name, None))
+        elif (field.one_to_one or field.many_to_one) and hasattr(field, "rel_class"):
+            value = smart_str(
+                getattr(obj, field.get_attname(), None), strings_only=True
+            )
         else:
             value = smart_str(getattr(obj, field.name, None))
     except ObjectDoesNotExist:
--- a/auditlog/models.py
+++ b/auditlog/models.py
@@ -2,7 +2,7 @@ import ast
 import json
 from copy import deepcopy
 from datetime import timezone
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Union
 
 from dateutil import parser
 from dateutil.tz import gettz
@@ -10,7 +10,11 @@ from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
 from django.contrib.contenttypes.models import ContentType
 from django.core import serializers
-from django.core.exceptions import FieldDoesNotExist
+from django.core.exceptions import (
+    FieldDoesNotExist,
+    ObjectDoesNotExist,
+    ValidationError,
+)
 from django.db import DEFAULT_DB_ALIAS, models
 from django.db.models import Q, QuerySet
 from django.utils import formats
@@ -492,6 +496,9 @@ class LogEntry(models.Model):
                             value = formats.localize(value)
                         except ValueError:
                             pass
+                    elif field_type in ["ForeignKey", "OneToOneField"]:
+                        value = self._get_changes_display_for_fk_field(field, value)
+
                     # check if length is longer than 140 and truncate with ellipsis
                     if len(value) > 140:
                         value = f"{value[:140]}..."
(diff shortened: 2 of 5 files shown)


auditlog/diff.py
```diff
@@ -67,6 +67,10 @@ def get_field_value(obj, field):
                 value = timezone.make_naive(value, timezone=timezone.utc)
         elif isinstance(field, JSONField):
             value = field.to_python(getattr(obj, field.name, None))
+        elif (field.one_to_one or field.many_to_one) and hasattr(field, "rel_class"):
+            value = smart_str(
+                getattr(obj, field.get_attname(), None), strings_only=True
+            )
```
sum-rock, line 73: Here is the primary fix.

auditlog/models.py
```diff
@@ -486,6 +493,31 @@ def changes_display_dict(self):
...
+        # changed between the time the LogEntry is created and this method is called.
+        except ValidationError:
+            return value
+        # Attempt to return the string representation of the object
+        try:
+            return smart_str(field.related_model.objects.get(pk=pk_value))
+        # ObjectDoesNotExist will be raised if the object was deleted.
+        except ObjectDoesNotExist:
+            return f"Deleted '{field.related_model.__name__}' ({value})"
+
```
rposborne, line 520: One consequence of this is that the related objects' str representation is based on the CURRENT state of that object, not the historical state. This is not... great when looking at the audit log of object A, it should not tell you the lifecycle of object B. 

There are probably ways around this by using the serialized data feature, but this has quite a few drawbacks as serialized data may not be compatible with the current code. 

One possible solution is storing a tuple in the changes dict `(pk, str)` but this changes the shape in other ways and represents a more complex breaking change. 

**or maybe having a datetime in a __str__ method is just an antipattern** (the root issue, on why we are here)

sum-rock: String representations are usually used as a way to have a human readable display for the object. I do not believe that `__str__` is intended to show state. Therefore showing the human readable form of which object was attached to this field on this object may not be so problematic. One could also argue that displaying the current version of the string representation for both the old and new relationships is a feature add because it's easier to compare. That said, I don't think people are likely to change the format of the returned `__str__` value that often. (Although losing the string representation when the object is deleted is unfortunate.)

auditlog_tests/tests.py
```diff
@@ -1551,6 +1551,142 @@ def test_when_field_doesnt_exist(self):
...
+        t2 = self.test_date + datetime.timedelta(days=20)
+        with freezegun.freeze_time(t2):
+            instance.related = two_simple
+            instance.save()
+
+        self.assertEqual(instance.history.all().count(), 2)
+        log_create = instance.history.filter(timestamp=t1).first()
+        log_update = instance.history.filter(timestamp=t2).first()
+        self.assertEqual(int(log_create.changes_dict["related"][1]), one_simple.id)
+        self.assertEqual(int(log_update.changes_dict["related"][1]), two_simple.id)
```
sum-rock, line 1687: Without the primary fix, this test case also fails. This seems pretty likely to be happening in other installations.

auditlog/models.py
```diff
@@ -486,6 +493,31 @@ def changes_display_dict(self):
             changes_display_dict[verbose_name] = values_display
         return changes_display_dict
 
+    def _get_changes_display_for_fk_field(
+        self, field: Union[models.ForeignKey, models.OneToOneField], value: Any
+    ) -> str:
+        """
+        :return: An object display string for a given FK value and the field to which it belongs
```
hramezani, line 500: I think this function does not return an object

hramezani:
I think this change is a breaking change as the patch is changing the representation of the related field. we need to add changelog for that and explain the change and the breaking change.

Also, It would be great if @alieh-rymasheuski can take a look.

hramezani:
Thanks @sum-rock for addressing the comments. As I mentioned in https://github.com/jazzband/django-auditlog/pull/420#issuecomment-1256233511 it is a breaking change and we can not include it in `2.2`.
we will include it in `3.0`

hramezani:
@sum-rock we are preparing `v3.0`. Could you please refresh this PR?

sum-rock:
@hramezani This branch is now refreshed

hramezani (approved):

Issue #426, opened 2022-09-15
[Enhancement] Use verbose name of field for display purpose


rptmat57:
A possible enhancement for the **admin pages** and the **display str** functions would be to try to find the verbose name of the field and use it (and even capitalize it):

```python
try:
    field = logentry.content_type.model_class()._meta.get_field(field).verbose_name.capitalize()
except:
    pass
```

Just a thought.

hramezani:
Thanks @rptmat57  for proposing this.
Could you please provide a picture of the change on the admin page?
Do you want to prepare a patch for your suggestion?

hramezani:
Closed by https://github.com/jazzband/django-auditlog/commit/d74c118834ac9f4f5dd76fd443fe905491fcd976

PR #428, opened 2022-09-15
Added verbose field name in admin

rptmat57:
Added verbose field name also using mapping name if defined (previously the mapping would not be used)

### DIFF
--- a/auditlog/mixins.py
+++ b/auditlog/mixins.py
@@ -2,12 +2,15 @@ import json
 
 from django import urls as urlresolvers
 from django.conf import settings
+from django.core.exceptions import FieldDoesNotExist
+from django.forms.utils import pretty_name
 from django.urls.exceptions import NoReverseMatch
 from django.utils.html import format_html, format_html_join
 from django.utils.safestring import mark_safe
 from django.utils.timezone import localtime
 
 from auditlog.models import LogEntry
+from auditlog.registry import auditlog
 
 MAX = 75
 
@@ -81,7 +84,7 @@ class LogEntryAdminMixin:
             msg.append("<table>")
             msg.append(self._format_header("#", "Field", "From", "To"))
             for i, (field, change) in enumerate(sorted(atom_changes.items()), 1):
-                value = [i, obj.verbose_name(field)] + (
+                value = [i, self.field_verbose_name(obj, field)] + (
                     ["***", "***"] if field == "password" else change
                 )
                 msg.append(self._format_line(*value))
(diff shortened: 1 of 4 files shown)


rptmat57:
implementation of #426

hramezani:
@rptmat57 do you want to fix the failing tests?

auditlog/models.py
```diff
@@ -480,12 +478,29 @@ def changes_display_dict(self):
...
+        model_fields = auditlog.get_model_fields(model._meta.model)
+        mapping_field_name = model_fields["mapping_fields"].get(field_name)
+        if mapping_field_name:
+            return mapping_field_name
+        else:
+            try:
+                field = model._meta.get_field(field_name)
+                verbose_name = getattr(field, "verbose_name", None)
+                return (
+                    verbose_name.lstrip().capitalize() if verbose_name else field_name
```
hramezani, line 498: I think we can remove the `capitalize` here to be backward compatible

auditlog/mixins.py
```diff
@@ -99,7 +101,7 @@ def msg(self, obj):
                     format_html(
                         "<tr><td>{}</td><td>{}</td><td>{}</td><td>{}</td></tr>",
                         i,
-                        field,
+                        obj.verbose_name(field),
```
hramezani, line 104: Do we have any tests for that?

hramezani:
@rptmat57 You need to add a changelog entry in https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md

auditlog/models.py
rptmat57: the idea here was to display field name similar to django-admin's "pretty_name" function.
thinking about it, it might be better to actually use that function rather than adding more logic.

auditlog/models.py
```diff
@@ -480,12 +479,28 @@ def changes_display_dict(self):
...
 
                     values_display.append(value)
-            verbose_name = model_fields["mapping_fields"].get(
-                field.name, getattr(field, "verbose_name", field.name)
-            )
+            verbose_name = self.verbose_name(field_name)
             changes_display_dict[verbose_name] = values_display
         return changes_display_dict
 
+    def verbose_name(self, field_name: str):
```
hramezani, line 486: I would say let's just move the logic to `verbose_name` and reuse it in `mixins.py` and do not change it.

rptmat57: alright then. if that's what you prefer!

auditlog/mixins.py
rptmat57: yes there is already a test for that ("test_changes_msg_m2m")

auditlog_tests/tests.py
```diff
@@ -1230,7 +1230,7 @@ def setUp(self):
 
     @property
     def latest_array_change(self):
-        return self.obj.history.latest().changes_display_dict["arrayfield"][1]
+        return self.obj.history.latest().changes_display_dict["Arrayfield"][1]
```
hramezani, line 1233:
```suggestion
        return self.obj.history.latest().changes_display_dict["arrayfield"][1]
```

hramezani (approved):
Thanks :+1:

PR #429, opened 2022-09-16
added fix for OneToOneRel error happening in case of Polymorphic model

rahulpr27git:


hramezani:
Thanks @rahulpr27git for this patch :+1: 
Please add a regression test and also add a changelog entry in [CHANGELOG.md](https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md)

tox.ini
```diff
@@ -21,6 +21,7 @@ deps =
     codecov
     freezegun
     psycopg2-binary==2.8.6
+    django-polymorphic
```
hramezani, line 24: Do we need `django-polymorphic` for testing? isn't it possible to add test without this?

rahulpr27git: yes. it's happening with a Polymorphic model so need this dependency else i wont be able to reproduce.

hramezani: I would take a look. it may be possible to reproduce the problem without `Polymorphic`. Otherwise, we can accept it.

rahulpr27git: Ok. thanks. 
this is currently became a major blocker for my project.

hramezani: btw, test still failing

rahulpr27git: where is this rdoc?
`pip install -rdocs/requirements.txt`

hramezani: Thanks @rahulpr27git.

Seems hard to reproduce the problem without `Polymorphic`. I would suggest just keeping the change and reverting all the changes related to the test. we will accept this change without tests. Because the fix itself is clear and I don't want to include another project just for one single test.

rahulpr27git: @hramezani please check again.

hramezani (approved):
LGTM :+1:

tox.ini
hramezani: Thanks @rahulpr27git

PR #432, opened 2022-09-21
Support search by custom USERNAME_FIELD

### LINKED ISSUE #431: Related Field got invalid lookup: username
- noticed while deleting logentires

alieh-rymasheuski:
Fixes #431

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - fix: Display `created` timestamp in server timezone ([#404](https://github.com/jazzband/django-auditlog/pull/404))
 - fix: Handle port in `remote_addr` ([#417](https://github.com/jazzband/django-auditlog/pull/417))
 - fix: Handle the error with AttributeError: 'OneToOneRel' error occur during a `PolymorphicModel` has relation with other models  ([#429](https://github.com/jazzband/django-auditlog/pull/429))
+- fix: Support search by custom USERNAME_FIELD ([#432](https://github.com/jazzband/django-auditlog/pull/432))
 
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks @alieh-rymasheuski for the quick fix.
Is it possible to add a test for this? Also change log is missing.

I think we can prepare `2.2` when we merge this.

alieh-rymasheuski:
@hramezani, the admin configuration is a tricky thing to test, especially given that we evaluate `LogEntryAdmin.search_fields` on import (so we are limited in our options of using `unittest.mock.patch` to tinker with the user model). If you have an idea how to test this properly without switching the user model for all tests to a non-default one, please try to write such test. Otherwise let's keep this change untested (together with the majority of the configuration of the admin).

hramezani:
> @hramezani, the admin configuration is a tricky thing to test, especially given that we evaluate `LogEntryAdmin.search_fields` on import (so we are limited in our options of using `unittest.mock.patch` to tinker with the user model). If you have an idea how to test this properly without switching the user model for all tests to a non-default one, please try to write such test. Otherwise let's keep this change untested (together with the majority of the configuration of the admin).
> 
> Added the change log - sorry for not doing it before.

@alieh-rymasheuski agree.

Question: Have you tested it locally?

@dev-fairshares Could you please confirm the fix?

alieh-rymasheuski:
@hramezani, I tested and double-tested this change, it works for me.

hramezani (approved):

hramezani:
Thanks @alieh-rymasheuski fix fixing and @dev-fairshares for checking :+1:

Issue #345, opened 2022-01-10
Logging access


simonkern:
Unlike other packages like [django-simple-history](https://github.com/jazzband/django-simple-history), the package name django-auditlog implies (at least to me) that all audit-related information can be logged using this package. Unfortunately, this is currently definitely not the case.

As of now Auditlog only handles three of the four CRUD operations. Only the following types of operations are currently covered:
- CREATE
- UPDATE
- DELETE

However, from time to time it's also necessary to log access to objects (READ). At least in the EU, where the GDPR is widely applied, it is the rule rather than the exception that access to PII must be logged. These logs are then viewed and reviewed as part of data protection audits. Nearly all of the data protection officers with whom we work require such a mechanism.

This could be implemented, e.g., by means of a signal and a Mixin for usage in DetailViews etc. I have a unpolished but working version of this. Would this be of interest for the project and is there a chance to get this sort of addition merged?

Thanks!

alieh-rymasheuski:
@simonkern, I allowed myself to edit the description of the issue to include your motivation you described in https://github.com/jazzband/django-auditlog/pull/436#issuecomment-1273694785. At least for me, that motivation was sufficient to change my opinion from "meh, who needs to log reads, that would bloat the log" to "this is a real-world requirement for real-world audits therefore it's valid". Please feel free to adjust the wording or even undo my edit completely if you will.

simonkern:
@alieh-rymasheuski thank you very much. It is completely fine for me and I really appreciate this. I should have thought of this myself.

I hope I was able to convinve the others as well. What do you suggest how we should proceed here?

Whether this gets merged at some point or not, I think it makes sense to provide a mechanism for FBVs either way. What approach concerning FBVs would you and the other maintainers suggest / prefer?

simonkern:
This has been merged in https://github.com/jazzband/django-auditlog/pull/436

hramezani:
Fixed in #436

PR #436, opened 2022-09-27
Added ACCESS action and enabled logging of object accesses

simonkern:
* adds Action ACCESS in addition to CREATE, UPDATE and DELETE
* Provides LogAccessDetailView that creates an ACCESS entry in auditlog when accessed

**Open**:
- [x] Doc / readme updates
- [x] tests


simonkern:
Related to #345

simonkern:
I have included tests and changed the docs. Please let me know what you think.

2ykwang:
In my opinion, there are limited ways to use the ACCESS action. for example, integration for DRF, support for function based views is needed.

hramezani:
Thanks @simonkern for the implementation.

IMHO, logging `ACCESS` shouldn't be added to this package because it's not a `change` and this package is designed to audit the changes on objects.

Here is the first line from readme

> django-auditlog (Auditlog) is a reusable app for Django that makes logging object changes a breeze.

simonkern:
Maybe I can convince you that the change fits very well to django-auditlog after all, if I explain the considerations that led to this PullRequest:

Unlike other packages like [django-simple-history](https://github.com/jazzband/django-simple-history), the package name django-auditlog implies (at least to me) that all audit-related information can be logged using this package. Unfortunately, this is currently definitely not the case.
At least in the EU, where the GDPR is widely applied, it is the rule rather than the exception that access to PII must be logged. These logs are then viewed and reviewed as part of data protection audits. Nearly all of the data protection officers with whom we work require such a mechanism.

Logging access really only requires minimal adjustments to the existing code base, so the effort required to enable logging access is manageable. We just need to add another action and provide mechanisms for logging that action. It didn't seem to make much sense to create a full-blown fork when minimal adjustments can cover the usecase described earlier.

At least 3 other people who read the underlying issue (#345) seem to have a similar need, or at least think extending the feature set is a good idea. Perhaps these people can comment briefly here (@radokristof, @mojicae and @rposborne).

In the end, I don't think it will hurt the package, but rather broaden its applicability.

However, I agree with you that another mechanism for FBVs should be provided, and the example integration with DRF should be shown.

auditlog/models.py
```diff
@@ -308,17 +308,20 @@ class Action:
...
 
-        The valid actions are :py:attr:`Action.CREATE`, :py:attr:`Action.UPDATE` and :py:attr:`Action.DELETE`.
+        The valid actions are :py:attr:`Action.CREATE`, :py:attr:`Action.UPDATE`,
+        :py:attr:`Action.DELETE` and :py:attr:`Action.ACCESS`.
         """
 
         CREATE = 0
         UPDATE = 1
         DELETE = 2
+        ACCESS = 5
```
alieh-rymasheuski, line 318: Why 5? :confused:

simonkern: That was just a random choice to avoid conflicts in case other people also forked auditlog and fiddled with its actions. I am happy to change it to whatever you prefer.

auditlog/migrations/0008_auto_20210826_1434.py
```diff
@@ -0,0 +1,28 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name="logentry",
+            name="action",
+            field=models.PositiveSmallIntegerField(
+                choices=[(0, "create"), (1, "update"), (2, "delete"), (5, "access")],
+                verbose_name="action",
+            ),
+        ),
+        migrations.AlterField(
```
hramezani, line 21: Do we need this migration?

auditlog/migrations/0012_merge_20220927_1126.py
```diff
@@ -0,0 +1,13 @@
+# Generated by Django 4.1.1 on 2022-09-27 09:26
```
hramezani, line 1: Why do we have this merge migration here?

simonkern: Between the time I applied the changes to my local branch and the time I opened this PR, there were some changes made to auditlog, that force me to create a merge migration. However, I can delete my initial migration and the merge migratiion and create a new one on top.

auditlog/migrations/0008_auto_20210826_1434.py
simonkern: See below, I will recreate the migrations.

auditlog/receivers.py
```diff
@@ -61,6 +61,22 @@ def log_delete(sender, instance, **kwargs):
...
 
 
+def log_access(sender, instance, **kwargs):
+    """
+    Signal receiver that creates a log entry when a model instance is accessed in a AccessLogDetailView.
+
+    Direct use is discouraged, connect your model through :py:func:`auditlog.registry.register` instead.
+    """
+    if instance.pk is not None:
+        changes = model_instance_diff(instance, instance)
```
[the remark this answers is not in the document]
2ykwang: https://github.com/s2k-digital/django-auditlog/blob/609ddb648dd25e3b4241fde66829d28897edbd39/auditlog/diff.py#L173-L174

expect to return `None` if no changes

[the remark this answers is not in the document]
hramezani: ah, yes. You are right. seems we don't need to call `model_instance_diff`. it is always `None`

auditlog/migrations/0012_merge_20220927_1126.py
hramezani: Please do it

auditlog_tests/tests.py
```diff
@@ -1797,3 +1798,29 @@ def test_serialize_related_with_kwargs(self):
...
+        new_count = qs.count()
+        self.assertEqual(new_count, old_count + 1)
+
+        log_entry = qs.latest()
+        self.assertEqual(int(log_entry.object_pk), self.obj.pk)
+        self.assertEqual(log_entry.actor, self.user)
+        self.assertEqual(log_entry.content_type, content_type)
+        self.assertEqual(
+            log_entry.action, LogEntry.Action.ACCESS, msg="Action is 'ACCESS'"
+        )
```
hramezani, line 1826: Please test the content of the `changes` field in this test.

auditlog/receivers.py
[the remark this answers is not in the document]
simonkern: `model_instance_diff` should in fact always be None and`json.dumps(None)` will give us `"null"`.  I've just put in `"null"` directly.

simonkern:
I've implemented all recommended changes and squashed the commits.

One point remains open: How to deal with FBVs? Provide a utility method that takes an object as a parameter and creates the coressponding access LogEntry?

auditlog_tests/tests.py
hramezani: @simonkern have you checked this comment?

simonkern: I've just received an email notification about this change request, but cannot find your comment quoted in the mail, so I'll reply here:

Yes I checked that comment and introduced a check:
https://github.com/s2k-digital/django-auditlog/blob/master/auditlog_tests/tests.py#L1827

simonkern:
Thanks @hramezani. I added a section for it, let me know if you want me to make any changes.

hramezani (dismissed):
LGTM :+1:

hramezani:
@alieh-rymasheuski Do you want to take another look?

alieh-rymasheuski (dismissed):

auditlog/receivers.py
```diff
@@ -61,6 +61,21 @@ def log_delete(sender, instance, **kwargs):
...
+    Signal receiver that creates a log entry when a model instance is accessed in a AccessLogDetailView.
+
+    Direct use is discouraged, connect your model through :py:func:`auditlog.registry.register` instead.
+    """
+    if instance.pk is not None:
+
+        LogEntry.objects.log_create(
+            instance,
+            action=LogEntry.Action.ACCESS,
+            changes="null",
```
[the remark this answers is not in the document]
simonkern: Hi @alieh-rymasheuski, if you take a look at the delete variant of the receiver, you'll see that changes is the result of
``` python
changes = model_instance_diff(instance, None)
```
see https://github.com/jazzband/django-auditlog/blob/master/auditlog/receivers.py#L55

In the context of access, this will always be None, as we pointed out above, see
https://github.com/jazzband/django-auditlog/pull/436#pullrequestreview-1137603666

The encoding into json will therefore return `'null'`, so we can skip all these calls and just put in `'null'` directly.
```
In [1]: import json

In [2]: json.dumps(None)
Out[2]: 'null'
```

hramezani:
@simonkern Could you please fix the conflict and refresh your PR. Then I can merge it.

hramezani (approved):

simonkern:
@hramezani done

PR #404, opened 2022-08-01
Display timestamps in server timezone

alieh-rymasheuski:
Resolves #403

PR #438, opened 2022-10-04
Remove unnecessary code in `created` method.

2ykwang:
related #404 

already run the same code in the Django
ref: https://github.com/django/django/blob/3.2/django/contrib/admin/utils.py#L381-L429

```python
# ...
elif isinstance(field, models.DateTimeField):
    return formats.localize(timezone.template_localtime(value))
# ...
```

### DIFF
--- a/auditlog/mixins.py
+++ b/auditlog/mixins.py
@@ -21,7 +21,6 @@ class LogEntryAdminMixin:
     def created(self, obj):
         return localtime(obj.timestamp)
 
-
     @admin.display(description="User")
     def user_url(self, obj):
         if obj.actor:
@@ -34,7 +33,6 @@ class LogEntryAdminMixin:
             return format_html('<a href="{}">{}</a>', link, obj.actor)
 
         return "system"
-
 
     @admin.display(description="Resource")
     def resource_url(self, obj):
@@ -50,7 +48,6 @@ class LogEntryAdminMixin:
                 '<a href="{}">{} - {}</a>', link, obj.content_type, obj.object_repr
             )
 
-
     @admin.display(description="Changes")
     def msg_short(self, obj):
         if obj.action == LogEntry.Action.DELETE:
@@ -62,7 +59,6 @@ class LogEntryAdminMixin:
             i = fields.rfind(" ", 0, MAX)
             fields = fields[:i] + " .."
         return "%d change%s: %s" % (len(changes), s, fields)
-
 
     @admin.display(description="Changes")
     def msg(self, obj):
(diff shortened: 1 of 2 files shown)


alieh-rymasheuski:
@2ykwang, the main part of the `created` method is a call to `localtime`. It matters when the server time is not UTC (it's set to NY time in my case).

These are the outputs of your version, the current version, and a compromise version which delegates the formatting to Django code:
```python
    def created(self, obj):
        return obj.timestamp
```
Created: `2022-10-03 16:09:59.723996+00:00`
```python
    def created(self, obj):
        return localtime(obj.timestamp).strftime("%Y-%m-%d %H:%M:%S")
```
Created: `2022-10-03 12:09:59`
```python
    def created(self, obj):
        return localtime(obj.timestamp)
```
Created: `2022-10-03 12:09:59.723996-04:00`

I would argue that precision to seconds and implicit server-local timezone are beneficial to the readability.

2ykwang:
@alieh-rymasheuski When the `created` method returns a `DateTimeField` object, it displays the data formatted according to Django settings. in my opinion, it felt awkward because the `LogEntryMixin.created` date format was different from other Django Admin date formats.

alieh-rymasheuski:
Agree. But please make sure to keep the `localtime` call in place - it is valuable for the case when the server timezone is different from UTC.

2ykwang:
@alieh-rymasheuski 
thank you I just edited the code.

alieh-rymasheuski (approved):

alieh-rymasheuski:
@2ykwang, please add a changelog entry as well.

@hramezani, would you like to take a look too?

2ykwang:
@alieh-rymasheuski done!

auditlog/mixins.py
```diff
@@ -16,11 +17,11 @@
 
 
 class LogEntryAdminMixin:
+    @admin.display(description="created")
```
hramezani, line 20:
```suggestion
    @admin.display(description="Created")
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 #### Fixes
 
+- fix: Remove unnecessary code in `created` method. ([#438](https://github.com/jazzband/django-auditlog/pull/438))
```
hramezani, line 10: I think we don't need to mention this here as it's not adding/changing a feature.
Please remove it.

hramezani:
Thanks @2ykwang for the patch and @alieh-rymasheuski for review!

Issue #443, opened 2022-10-19
Config to skip auditlog during imports


robinharms:
It would be great if it was possible to have a simple setting to disable logging for imports. (Essentially listening for raw=True)

Preferably a global setting. If this is something you world consider, I would be happy to make a patch.

robinharms:
I've done the patch already, but can't seem to run the tests. :)
Sorry for asking dumb questions, executing runtests.py doesn't seem to work for me.

hramezani:
Hi @robinharms,

How do you run your tests locally? have to tried to run the tests with `tox`?

robinharms:
Actually no I hadn't, but when running the tests with tox I get the same error eventually:

  File "/Users/robin/PycharmProjects/django-auditlog/.tox/py39-django40/lib/python3.9/site-packages/django/dispatch/dispatcher.py", line 85, in connect
    raise TypeError("Signal receivers must be callable.")
TypeError: Signal receivers must be callable

hramezani:
Did you get this error in fresh master as well?

robinharms:
Yes, and I use the latest

hramezani:
It might be related to your tox environment. Are you running your test in the console by running for example `tox -e py39-django40` command?

robinharms:
Yeah I did.

Sure it has nothing to do with the changed signal handlers?

  File "/Users/robin/PycharmProjects/django-auditlog/.tox/py39-django40/lib/python3.9/site-packages/django/db/models/signals.py", line 27, in connect
    self._lazy_method(
  File "/Users/robin/PycharmProjects/django-auditlog/.tox/py39-django40/lib/python3.9/site-packages/django/db/models/signals.py", line 24, in _lazy_method
    return partial_method(sender)
  File "/Users/robin/PycharmProjects/django-auditlog/.tox/py39-django40/lib/python3.9/site-packages/django/dispatch/dispatcher.py", line 85, in connect
    raise TypeError("Signal receivers must be callable.")
TypeError: Signal receivers must be callable.

robinharms:
Wait forget about this, It's another package wrapping this :)
Sorry to trouble you!

I'll recheck the tests and come back with a PR for the changes I made.

hramezani:
Closed in https://github.com/jazzband/django-auditlog/commit/8fe776ae45964d20bf113d5bff9287398f89480a

PR #446, opened 2022-11-02
Option to disable logging on raw save and via context manager

robinharms:
Related to this issue: #443 

Since m2ms don't have the kwarg raw, I've added a context manager that can be used in situations when things shouldn't be logged.

My use-case is essentially large imports on a live server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - feat: Add `serialized_data` field on `LogEntry` model. ([#412](https://github.com/jazzband/django-auditlog/pull/412))
 - feat: Display the field name as it would be displayed in Django Admin or use `mapping_field` if available [#428](https://github.com/jazzband/django-auditlog/pull/428)
+- feat: New context manager `disable_auditlog` to turn off logging and a setting `AUDITLOG_DISABLE_ON_RAW_SAVE`
+  to disable it during raw-save operations like loaddata. [#446](https://github.com/jazzband/django-auditlog/pull/446)
 
 #### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - feat: Add `serialized_data` field on `LogEntry` model. ([#412](https://github.com/jazzband/django-auditlog/pull/412))
 - feat: Display the field name as it would be displayed in Django Admin or use `mapping_field` if available [#428](https://github.com/jazzband/django-auditlog/pull/428)
-- feat: New context manager `disable_auditlog` to turn off logging and a setting `AUDITLOG_DISABLE_ON_RAW_SAVE`
+- feat: New context manager `disable_auditlog` to turn off logging and a new setting `AUDITLOG_DISABLE_ON_RAW_SAVE`
   to disable it during raw-save operations like loaddata. [#446](https://github.com/jazzband/django-auditlog/pull/446)
 
 #### Fixes
--- a/auditlog/conf.py
+++ b/auditlog/conf.py
@@ -15,3 +15,8 @@ settings.AUDITLOG_EXCLUDE_TRACKING_MODELS = getattr(
 settings.AUDITLOG_INCLUDE_TRACKING_MODELS = getattr(
     settings, "AUDITLOG_INCLUDE_TRACKING_MODELS", ()
 )
+
+# Disable on raw save to avoid logging imports and similar
+settings.AUDITLOG_DISABLE_ON_RAW_SAVE = getattr(
+    settings, "AUDITLOG_DISABLE_ON_RAW_SAVE", False
+)
--- a/auditlog/context.py
+++ b/auditlog/context.py
@@ -64,3 +64,15 @@ def _set_actor(user, sender, instance, signal_duid, **kwargs):
             instance.actor = user
 
         instance.remote_addr = auditlog["remote_addr"]
+
+
+@contextlib.contextmanager
+def disable_auditlog():
+    threadlocal.auditlog_disabled = True
+    try:
+        yield
+    finally:
+        try:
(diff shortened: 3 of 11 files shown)


robinharms:
Btw I haven't added anything in the docs yet. If you want to merge I'll fix that of course.

auditlog/receivers.py
```diff
@@ -1,16 +1,39 @@
...
+    - 'auditlog_disabled' from threadlocal is true
+    - raw = True and AUDITLOG_DISABLE_ON_RAW_SAVE is True
+    """
+
+    @wraps(signal_handler)
+    def wrapper(*args, **kwargs):
+        disable = getattr(threadlocal, "auditlog_disabled", False)
+        if not disable:
+            disable = kwargs.get("raw") and settings.AUDITLOG_DISABLE_ON_RAW_SAVE
+        kwargs.setdefault("disable", disable)
```
alieh-rymasheuski, line 23: _(neither supporting nor opposing the PR as a whole)_

Instead of passing `disable` flag to the signal handler, can we instead just skip calling the signal handler if we want to "disable" it?
```python
disable = getattr(threadlocal, "auditlog_disabled", False) or kwargs.get("raw") and settings.AUDITLOG_DISABLE_ON_RAW_SAVE)
if not disable:
    signal_handler(*args, **kwargs)
```

robinharms: You're absolutely right, I'll fix that right away.

auditlog/conf.py
```diff
@@ -15,3 +15,8 @@
 settings.AUDITLOG_INCLUDE_TRACKING_MODELS = getattr(
     settings, "AUDITLOG_INCLUDE_TRACKING_MODELS", ()
 )
+
+# Disable on raw save to avoid logging imports and similar
+settings.AUDITLOG_DISABLE_ON_RAW_SAVE = getattr(
+    settings, "AUDITLOG_DISABLE_ON_RAW_SAVE", False
```
hramezani, line 21: Isn't it a change in the current behavior? Currently django-auditlog logs the import.
I would say the default value should be `True` to keep the old behavior and don't break it

auditlog/context.py
```diff
@@ -64,3 +64,15 @@ def _set_actor(user, sender, instance, signal_duid, **kwargs):
...
+@contextlib.contextmanager
+def disable_auditlog():
+    threadlocal.auditlog_disabled = True
+    try:
+        yield
+    finally:
+        try:
+            del threadlocal.auditlog_disabled
+        except AttributeError:
+            pass
```
hramezani, line 78: it would be great to have to test to cover this line

auditlog/registry.py
```diff
@@ -266,7 +266,8 @@ def register_from_settings(self):
         """
         if not isinstance(settings.AUDITLOG_INCLUDE_ALL_MODELS, bool):
             raise TypeError("Setting 'AUDITLOG_INCLUDE_ALL_MODELS' must be a boolean")
-
+        if not isinstance(settings.AUDITLOG_DISABLE_ON_RAW_SAVE, bool):
+            raise TypeError("Setting 'AUDITLOG_DISABLE_ON_RAW_SAVE' must be a boolean")
```
hramezani, line 270: We need to have a test for this line. something like https://github.com/jazzband/django-auditlog/blob/aa6d977f8b8b070690f9a168f2eddaa93a65df70/auditlog_tests/tests.py#L1043-L1047

hramezani:
Thanks @robinharms for this change. In general it LGTM :+1: 
I left some comments.
Please remember adding:

- changelog entry
- documentation for this new setting and context manager

auditlog/context.py
robinharms: The attribute error will never happen unless it's wrapped twice for some reason.
Deleting the attribute is covered here: https://github.com/jazzband/django-auditlog/pull/446/files#diff-eb9581b1c35dba2bdbfcf71d8c389814416cdbe08a5dadadc0ce3ab4520b61b2R1846

Is that okay?

auditlog/registry.py
robinharms: Good point, I've fixed that

auditlog/context.py
hramezani: Ok, let's have it like this

robinharms:
All done!

Note that I assumed the next release was 2.2.0 so I used that in the docs.

Sorry about the initial bad code ;)

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - feat: Add `serialized_data` field on `LogEntry` model. ([#412](https://github.com/jazzband/django-auditlog/pull/412))
 - feat: Display the field name as it would be displayed in Django Admin or use `mapping_field` if available [#428](https://github.com/jazzband/django-auditlog/pull/428)
+- feat: New context manager `disable_auditlog` to turn off logging and a setting `AUDITLOG_DISABLE_ON_RAW_SAVE`
```
hramezani, line 7:
```suggestion
- feat: New context manager `disable_auditlog` to turn off logging and a new setting `AUDITLOG_DISABLE_ON_RAW_SAVE`
```

hramezani:
> All done!
> 
> Note that I assumed the next release was 2.2.0 so I used that in the docs.
> 
> Sorry about the initial bad code ;)

Thanks for the update. I left a small suggestion.
Also, we have some syntax problems in the new docs added https://github.com/jazzband/django-auditlog/actions/runs/3385622412/jobs/5624027992. Please fix it as well.

robinharms:
Ah, missed the :: marker. Sorry

hramezani (approved):
LGTM :+1:

hramezani:
Thanks @robinharms for your effort!
Let's don't merge it for now, to let the other people review it.
I will merge it in next couple of days.

PR #449, opened 2022-11-14
Make log entries read-only in the admin

alieh-rymasheuski:
This is a follow-up fix after a real-world staff user deleted a few auditlog entries through the admin by mistake.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changes
 
+#### Fixes
+- fix: Make log entries read-only in the admin. ([#449](https://github.com/jazzband/django-auditlog/pull/449))
+
 ## 2.2.0 (2022-10-07)
 
 #### Improvements
--- a/auditlog/admin.py
+++ b/auditlog/admin.py
@@ -26,5 +26,10 @@ class LogEntryAdmin(admin.ModelAdmin, LogEntryAdminMixin):
     ]
 
     def has_add_permission(self, request):
-        # As audit admin doesn't allow log creation from admin
+        return False
+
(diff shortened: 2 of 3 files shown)


hramezani:
Thanks @alieh-rymasheuski for this. two things:

- I think it's a breaking change. I think we have to make it configurable or mention it as a breaking change.(note: next release will be a `3.0` which is a major release, so we can consider this one as well.
- Test are failing please adopt them

alieh-rymasheuski:
> I think it's a breaking change

I think it's a bugfix. This is *audit* log, it shouldn't be possible to alter it with a simple button in the admin.

hramezani:
@alieh-rymasheuski Do you have a plant to fix the failing test?

alieh-rymasheuski:
@hramezani, done. :)

hramezani (approved):

madomdy:
I'm afraid with that, we can't delete any record from django admin. 😬😄

alieh-rymasheuski:
@madomdy, do you mean that this change causes django admin to disallow deleting auditlogged objects? I am confident that _some_ objects still can be deleted - in fact, I removed one this morning. :smile: 

Could you please elaborate on your issue? I'm happy to help and fix the bug that I may have introduced.

image

madomdy:
@alieh-rymasheuski happily! Here is the thing:
image (11)

And this happens on whatever object of whatever model connected to the audit log. I am just a bit confused how come others do not encounter it.

alieh-rymasheuski:
@madomdy, I registered [an issue](https://github.com/jazzband/django-auditlog/issues/493) to continue this discussion there.

PR #450, opened 2022-11-14
Replace `pkg_resources` with `importlib` solution

hramezani:


hramezani:
@alieh-rymasheuski Please take a look when you have time.

alieh-rymasheuski (approved):

PR #464, opened 2022-11-29
fix changes_display_dict to display added/deleted objects for m2m updates

a-cesari:
In one of my projects I needed to use the changes_display_dict function in a Django template to display changes of a m2m field.
Apparently for such field the `changes_display_dict` function is not storing the added/deleted object while the `changes_dict` function does. The changes_display_dict function just store a list like this: `['type', 'operation', 'objects']` .
See screenshots attached. 

changes_dict
changes_display_dict

I slightly modified the `changes_display_dict` function to also store the list of added/removed objects. The result is the following:

changes_display_dict_after

I hope you could consider merging this if appropriate. Happy to implement any changes/improvements you might suggest on my current proposal.

### DIFF
--- a/auditlog/models.py
+++ b/auditlog/models.py
@@ -464,6 +464,16 @@ class LogEntry(models.Model):
                 except AttributeError:
                     # if the field is a relationship it has no internal type and exclude it
                     continue
+                
+                if field_type == "ManyToManyField":
+                    values_display = values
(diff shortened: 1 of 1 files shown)


alieh-rymasheuski:
@a-cesari, thanks for your patch! I indeed missed `changes_display_dict` logic when implementing m2m tracking because it's a feature which I don't use myself.

Could you please cover the added logic with a test?

dengeeks (approved):

Issue #469, opened 2022-12-11
Correlation ID for records


aqeelat:
Oftentimes, when a change is made, it is made across multiple models/instances.

For example: when an order is paid:
1. update the order model
2. create a transaction in the ledger

A nice way to track this is to assign a correlation id to the request. We're currently doing this using [django-cid](https://pypi.org/project/django-cid/).
We:
1. define a cid header. if the header doesn't exist in the request, create a random UUID.
2. return the cid in get_additional_data
3. override auditlog admin to display the cid in the changelist

It's really helpful and I would love to see this being officially supported.
I'm willing to open a PR for this feature.

aqeelat:
This is how we have it:
<img width="1271" alt="Screen Shot 2022-12-11 at 10 42 41 PM" src="https://user-images.githubusercontent.com/5212744/206925450-e054552e-90b8-461c-a1ee-7b0354a94412.png">

hramezani:
Thanks @aqeelat for reporting this.
Agree with you, It is a nice feature to have.
I would like to have it simple I mean we can include the header in the request and then save it when we write the log. I think we can have a separate model field for `Correlation ID`. Also, I think we don't need to add a `django-cid` as a new dependency and we can generate the correlation_id by `uuid`.
What do you think?

aqeelat:
I agree that it should be a separate model field.One major factor we need to consider is the ability to customize the header name. In one project we use, we get the correlation id from cloudflare in a header caller cf-ray, and in another project we get it from datadog in a header caller trace-id.I don’t mind not adding django-cid as a dependency, as long as we do it in a way that’s not very similar. How about we include the package within this repo and then sync it every few months? (I don’t really like this idea though)Sent from my iPhoneOn Dec 12, 2022, at 11:05 AM, Hasan Ramezani ***@***.***> wrote:﻿
Thanks @aqeelat for reporting this.
Agree with you, It is a nice feature to have.
I would like to have it simple I mean we can include the header in the request and then save it when we write the log. I think we can have a separate model field for Correlation ID. Also, I think we don't need to add a django-cid as a new dependency and we can generate the correlation_id by uuid.
What do you think?

aqeelat:
The thing is, django-cid has more features that are not needed in this project but are helpful overall.
For example, it has a context processor, a cursor wrapper, and a middleware.

[...]

PR #475, opened 2022-12-14
Draft implementation of cid

aqeelat:
Related to #469 

TODO:

- [ ] a proper code review
- [ ] admin integration
- [ ] tests
- [ ] model migration
- [ ] documentation
- [ ] changelog

### DIFF
--- a/auditlog/cid.py
+++ b/auditlog/cid.py
@@ -1,0 +1,88 @@
+"""
+This code is heavily influenced by https://github.com/Polyconseil/django-cid
+
+Copyright (c) 2016, Snowball One
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+list of conditions and the following disclaimer.
+* Redistributions in binary form must reproduce the above copyright notice,
+this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @aqeelat for the initial implementation. here is my thought:

- I was thinking about something simple. having a setting like `AUDITLOG_CID_HEADER` and then fetching the value of the header and storing it in DB.
- Do you think that we need to generate CID? what do you think @alieh-rymasheuski 

BTW, nice usage of `ContextVar`. it would be great to change the existing implementation and rewrite it by `ContextVar`. for sure in a separate PR.

aqeelat:
@hramezani 
How about this:
1. we work on the core requirements (header to contextvar, get from contextvar, admin, testing, docs).
2. If the users want interoperability with django-cid or any other tool, they can:
    a. disable the setter (maybe we can even put this in a separate middleware that can be [disabled at runtime](https://docs.djangoproject.com/en/4.1/topics/http/middleware/#marking-middleware-as-unused))
    b. change the getter function
3. debate the generation functionality back in #469 to not tie up this feature

We can even reduce the number of configs by tieing point 2a with 2b so that if the user overrides the getter function, we disable the setter (either by a regular if statement or by disabling the middleware).

What do you think?

Or more specifically:
1. should I get going on the other PR?
2. should I create another middleware or is a regular if-statement enough? (efficiency vs simplicity)

Issue #199, opened 2019-01-14
Use `default` instead of `auto_now_add`?


crazyscientist:
Hi there,

this is more like a question.

I'm looking into the possibilities of importing old data (that will be needed when going productive with the app) into the auditlog's model and was wondering if having a default for the `timestamp` instead of the `auto_now_add` would be a good idea?

PR #200, opened 2019-01-16
Make timestamp of LogEntry overwritable

crazyscientist:
* Replaced `auto_now_add=True` with `default=timezone.now`
* Added two testcases
* Added migration

This is very handy in order to import historic data from another source.

This relates to #199

PR #478, opened 2022-12-15
feat: Make timestamp in LogEntry overwritable

aqeelat:
Fixes https://github.com/jazzband/django-auditlog/issues/199
A replacement PR for #200 since the repo in that PR was deleted.


auditlog/models.py
```diff
@@ -355,7 +356,10 @@ class Action:
         blank=True, null=True, verbose_name=_("remote address")
     )
     timestamp = models.DateTimeField(
-        db_index=True, auto_now_add=True, verbose_name=_("timestamp")
+        default=django_timezone.now,
+        db_index=True,
+        auto_now_add=True,
```
hramezani, line 361: I think you need to remove `auto_now_add` to fix the error
```
auditlog.LogEntry.timestamp: (fields.E160) The options auto_now, auto_now_add, and default are mutually exclusive. Only one of these options may be present.
```

hramezani (approved):

hramezani:
Thanks for update @aqeelat

PR #479, opened 2022-12-15
Using changes_dict in admin list

### LINKED ISSUE #477: TypeError object of type 'NoneType' has no len() when changes are empty
The admin logentry list throws a `TypeError object of type 'NoneType' has no len()` when json.loads() return None
I think this comes from https://github.com/jazzband/django-auditlog/blob/703e3e4ba6ba7127cbd6a1bbf2597504089ba78c/auditlog/mixins.py#L58
where:
```python
changes = json.loads(obj.changes)
s = "" if len(changes) == 1 else "s"
        fields = ", ".join(changes.keys())
```

this is related (if not the exact same thing) to #471 

I think a fix would be as simple as replacing `json.loads()` by `obj.changes_dict`

rptmat57:
Using changes_dict to fix potential TypeError when changes are None in admin
Fixes #477

### DIFF
--- a/auditlog/mixins.py
+++ b/auditlog/mixins.py
@@ -1,5 +1,3 @@
-import json
-
 from django import urls as urlresolvers
 from django.conf import settings
 from django.contrib import admin
--- a/auditlog/mixins.py
+++ b/auditlog/mixins.py
@@ -54,7 +54,7 @@ class LogEntryAdminMixin:
     def msg_short(self, obj):
         if obj.action in [LogEntry.Action.DELETE, LogEntry.Action.ACCESS]:
             return ""  # delete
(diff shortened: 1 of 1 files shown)


hramezani (approved):

hramezani:
Thanks @rptmat57 :+1:

rptmat57:
I noticed even though this accepted and closed a month and a half ago, it wasn't included in the latest release.
In fact the last release only had a small bug fix. Is that because of the upcoming major 3.0.0 release?

hramezani:
@rptmat57 Yes exactly. we are preparing `v3` which is a major release

rptmat57:
ok thanks for responding so quickly! appreciate it

PR #481, opened 2022-12-18
Correlation ID - Part 1

aqeelat:
feat: Added support for storing correlation ids passed to a custom header


hramezani:
@aqeelat Thanks for this. this is a good starting point :+1: 

Please remember to rebase the patch and re-run the `makemigrations` command (because I have merged https://github.com/jazzband/django-auditlog/pull/478)

I will take a look later this week

aqeelat:
@hramezani updated.

auditlog/admin.py
```diff
@@ -34,3 +41,7 @@ def has_change_permission(self, request, obj=None):
 
     def has_delete_permission(self, request, obj=None):
         return False
+
+    def get_queryset(self, request):
```
hramezani, line 45: Why do we need this here?

auditlog/cid.py
```diff
@@ -0,0 +1,35 @@
...
+
+    if header and request and header in request.headers or header in request.META:
+        cid = request.headers.get(header)
+
+    # in theory, we shouldn't have to set the cid back to None because each request should be in a new thread.
+    # however, this was causing some tests to fail.
+    correlation_id.set(cid)
+
+
+def _get_cid() -> Union[str, None]:
```
hramezani, line 23:
```suggestion
def _get_cid() -> Optional[str]:
```

auditlog/cid.py
```diff
@@ -0,0 +1,35 @@
...
+    return correlation_id.get()
+
+
+def get_cid() -> Union[str, None]:
```
hramezani, line 27:
```suggestion
def get_cid() -> Optional[str]:
```

auditlog_tests/tests.py
```diff
@@ -1332,6 +1360,23 @@ def test_created_timezone(self):
...
                 self.assertEqual(created.strftime("%Y-%m-%d %H:%M:%S"), timestamp)
 
+    def test_cid(self):
+        self.client.force_login(self.user)
+        expected_response = '<a href="/admin/auditlog/logentry/?cid=123" '
+        expected_response += (
+            'title="Click to filter by records with this correlation id">'
+        )
+        expected_response += "123"
+        expected_response += "</a>"
```
hramezani, line 1370:
```suggestion
        expected_response = (
            '<a href="/admin/auditlog/logentry/?cid=123" '
            'title="Click to filter by records with this correlation id">123</a>'
        )
```

auditlog/mixins.py
```diff
@@ -112,6 +116,15 @@ def msg(self, obj):
 
         return mark_safe("".join(msg))
 
+    @admin.display(description="Correlation ID")
+    def cid_url(self, obj):
+        cid = obj.cid
+        if cid:
+            url = self._add_query_parameter("cid", cid)
+            return format_html(
+                '<a href="{}" title="{}">{}</a>', url, self.CID_TITLE, cid
```
hramezani, line 125: Please use `f-string` here

hramezani:
Thanks @aqeelat for the update.
I left some comments.

Please add documentation for the new settings and also add tests for checking the saved cid in the model.

auditlog/admin.py
aqeelat: Because we need to access the request in order to append the cid to it for filtering

auditlog/mixins.py
aqeelat: using `f-string` here defeats the purpose of using `format_html`

aqeelat:
@hramezani done

auditlog/cid.py
```diff
@@ -0,0 +1,60 @@
...
+    :return: None
+    """
+    cid = None
+    header = settings.AUDITLOG_CID_HEADER
+
+    if header and request and header in request.headers or header in request.META:
+        cid = request.headers.get(header)
+
+    # in theory, we shouldn't have to set the cid back to None because each request should be in a new thread.
+    # however, this was causing some tests to fail.
```
hramezani, line 28: Didn't get this comment here. do we really set cid here to None? what if the cid present in header

docs/source/usage.rst
```diff
@@ -234,6 +244,17 @@ Disables logging during raw save. (I.e. for instance using loaddata)
 
 .. versionadded:: 2.2.0
 
+**AUDITLOG_CID_HEADER**
+
+The request header containing the Correlation ID value to use in all log entries created as a result of the request.
+
+.. versionadded:: 2.3.0
```
hramezani, line 251: The next release would be a major one. we have some pending PR which are breaking changes and I want to include them in next release
```suggestion
.. versionadded:: 3.0.0
```

docs/source/usage.rst
```diff
@@ -234,6 +244,17 @@ Disables logging during raw save. (I.e. for instance using loaddata)
...
+
+**AUDITLOG_CID_GETTER**
+
+The function to use to retrieve the Correlation ID.
+
+.. versionadded:: 2.3.0
```
hramezani, line 257:
```suggestion
.. versionadded:: 3.0.0
```

docs/source/usage.rst
```diff
@@ -234,6 +244,17 @@ Disables logging during raw save. (I.e. for instance using loaddata)
...
 
+**AUDITLOG_CID_HEADER**
+
+The request header containing the Correlation ID value to use in all log entries created as a result of the request.
+
+.. versionadded:: 2.3.0
+
+**AUDITLOG_CID_GETTER**
+
+The function to use to retrieve the Correlation ID.
```
hramezani, line 255: Please mention here that the value of this setting can be a callable or a str import path

hramezani:
Thanks @aqeelat for updating the patch.
I've left some comments. Please address them.

Please also add test to proved saved cid value in model

auditlog/cid.py
aqeelat: The idea is that I had to move setting the contextvar outside of the if statement. Which means that we'll be explicitly setting cid to None if the cid is not in the header.

Django requests are multi threaded but the tests are not.

hramezani: Yeah, I see.  but I didn't get the explanation from your comment.
You may need to phrase it better.

docs/source/usage.rst
aqeelat: Updated

aqeelat:
> Thanks @aqeelat for updating the patch.
> 
> I've left some comments. Please address them.
> 
> 
> 
> Please also add test to proved saved cid value in model
> 
> 

I added the test already. It should be in the same middleware test. I compared the value in the history with the expected value.

auditlog/cid.py
aqeelat: How about:
"Ideally, this line should be nested inside the if statement. However, because the tests do not run requests in multiple threads, we have to always set the value of the cid, even if the request does not have the header present, in which case it will be set to None"

hramezani: Great!

hramezani, line 28:
```suggestion
    # Ideally, this line should be nested inside the if statement. However, because the tests
    # do not run requests in multiple threads, we have to always set the value of the cid,
    # even if the request does not have the header present, in which case it will be set to None
```

hramezani:
```suggestion
    # Ideally, this line should be nested inside the if statement. However, because
    # the tests do not run requests in multiple threads, we have to always set the
    # value of the cid, even if the request does not have the header present, in which
    # case it will be set to None
```

hramezani:
I think everything is done. you just need to update the comment in `cid.py`. Then we are good to go

auditlog/cid.py
aqeelat: pushed

aqeelat:
@hramezani all is done now. LMK if you need anything else.

hramezani (approved):

PR #484, opened 2022-12-22
Always call set_actor

aqeelat:
My take on #279

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,20 +1,21 @@
 # Changes
 
+## Next Release
+
 #### Breaking Changes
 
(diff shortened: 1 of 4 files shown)


aqeelat:
@hramezani I guess a better way to do it is to create a function called `get_metadata` that gets called in `LogEntry.objects.log_create`. This function gets actor, cid, and remote_addr. This way, we wouldn't need a context manager or a pre_save signal.

hramezani:
@aqeelat Thanks for this patch :+1: 
Could you please refresh this patch?

hramezani:
@aqeelat Could you please fix the failing tests?

hramezani (approved):

PR #489, opened 2022-12-29
Fix repr of a json field in field changes

### LINKED ISSUE #476: Changes logged for JSON-field aren't valid JSON
Hello everyone,

suppose I have the following model:
```
from django.db import models

class Foo(models.Model):
    bar = models.JSONField()


auditlog.register(Foo)
```

Now I create a new instance and check for the last LogEntry:
```
Foo.objects.create(bar={"baz":1})
entry = LogEntry.objects.latest()
bar_changes = entry.changes_dict.get("bar")
new_bar = bar_changes[1]
print(new_bar)
```

This outputs:
```
{'baz': 1}
```

I would expect the log entries for JsonFields to be a valid string to feed into `json.loads` however, the double quotes around the keys get converted into single quotes which is not allowed for json. I did some digging into where this conversion happens but I didn't really come to a conclusion.

Is this the desired behaviour or is it a bug?

CleitonDeLima:
Resolve #476 

My doubt is related to the [`sort_keys=True`](https://github.com/jazzband/django-auditlog/compare/master...CleitonDeLima:fix/476?expand=1#diff-575428ad5f7c4dfa79028063532473346bd6e1fb20894b2aa60053fad4e96cdfR78) of the dumps, I believe it is worth keeping for comparison reasons.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 #### Improvements
 
+- Changes the view when it has changes in fields `JSONField`. The `JSONField.encoder` is assigned to `json.dumps`. ([#489](https://github.com/jazzband/django-auditlog/pull/489))
 - feat: Added support for Correlation ID. ([#481](https://github.com/jazzband/django-auditlog/pull/481))
 - feat: Added pre-log and post-log signals. ([#483](https://github.com/jazzband/django-auditlog/pull/483))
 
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @CleitonDeLima :+1: 

Please add a changelog entry in `Improvements` section of https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md

hramezani (approved):

PR #495, opened 2023-01-12
feat: give users the option to run the json migration asyncly

aqeelat:


hramezani:
Thanks @aqeelat for your effort!

I assume that users who want to migrate normally, just run `migrate` command and see a `changes_text` in the model and everything else is like before. right?

The general approach is ok for me we need to document it.

@alieh-rymasheuski do you have any idea here?

auditlog/management/commands/auditlogmigratejson.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "-d",
+            "--database",
+            default=None,
+            help="If provided, we will use native db operations.",
+            dest="db",
+            type=str,
+            choices=["postgres", "mysql", "oracle", "sqllite3"],
```
hramezani, line 19: I think we don't support `sqllite3`

alieh-rymasheuski: Also it's `sqlite3`, single `l`.

aqeelat:
@hramezani would you be able to write the documentation for it? Maybe in a "migration to version 3 guide" in the docs, or as a readme section.

I could write the bullet points as a comment to this PR but I'm not familiar with reStructuredText enough to write a good document.

Also, how do you recommend we write the tests for this pr?

auditlog/management/commands/auditlogmigratejson.py
aqeelat: Oh. I saw it included in https://github.com/jazzband/django-auditlog/pull/407#issuecomment-1205656698 and I thought it was supported.

hramezani:
> @hramezani would you be able to write the documentation for it? Maybe in a "migration to version 3 guide" in the docs, or as a readme section.
> 
> I could write the bullet points as a comment to this PR but I'm not familiar with reStructuredText enough to write a good document.
> 

Ok, Please write them down in this PR and I will take care of documentation later

> Also, how do you recommend we write the tests for this pr?

- We need some tests for the new `changes_func` 
- We need to test the new management command as well.

Please consider to add them in separate class to be able to remove them easily later or add some note regarding the removal

aqeelat:
I added a few tests but I'll come back later to polish them and add the changelog entry

aqeelat:
Now that I think of it, should I attach this PR to the current changelog entry?
Also, please review the comments I added to the pr

auditlog/management/commands/auditlogmigratejson.py
```diff
@@ -0,0 +1,110 @@
...
+            default=None,
+            help="If provided, we will use native db operations.",
+            dest="db",
+            type=str,
+            choices=["postgres", "mysql", "oracle"],
+        )
+        parser.add_argument(
+            "-b",
+            "--bactch-size",
+            default=None,
```
hramezani, line 24: isn't it better to have a default batch size?
I assume if the batch size is `None` it will process all the records in one shot am afraid it put pressure on databases with a lot of records.

auditlog/management/commands/auditlogmigratejson.py
```diff
@@ -0,0 +1,110 @@
...
+            help="Split the migration into multiple batches",
```
hramezani, line 25: Please mention that this works just in case `-d` is not provided.

auditlog/migrations/0015_alter_logentry_changes.py
```diff
@@ -1,18 +1,45 @@
...
     ]
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("auditlog", "0014_logentry_cid"),
+    ]
+
+    atomic = not settings.AUDITLOG_TWO_STEP_MIGRATION
```
hramezani, line 43: What is the usage of this line here?

auditlog/apps.py
```diff
@@ -11,3 +11,7 @@ def ready(self):
         from auditlog.registry import auditlog
 
         auditlog.register_from_settings()
+
+        from auditlog import models
+
+        models.changes_func = models._changes_func()
```
hramezani, line 17: I would move it to `auditlog/models.py`

hramezani:
@aqeelat  Thanks for your update.

> Now that I think of it, should I attach this PR to the current changelog entry?

Yes

> Also, please review the comments I added to the pr

It's because of your `get_logs` query. at the beginning we have L1, L2 and L3. the migration uses the first page of get_log which is L1. then in next loop, get_logs query result is L2 and L3(because L1 is migrated). So in the second loop migration picks page 2 from L2 and L3 wich is L3.
For fixing the problem I think you need to don't use paginator and always pick the first `batch_size` items of `get_logs`

auditlog/apps.py
aqeelat: When I had it in the models, it was throwing an error saying that the conf is not found. Do you have any suggestions on how to fix it?

aqeelat:
> It's because of your `get_logs` query. at the beginning we have L1, L2 and L3. the migration uses the first page of get_log which is L1. then in next loop, get_logs query result is L2 and L3(because L1 is migrated). So in the second loop migration picks page 2 from L2 and L3 wich is L3. For fixing the problem I think you need to don't use paginator and always pick the first `batch_size` items of `get_logs`

Good catch. I didn't know that the paginator reevaluates the qs in each iteration. I replaced it with a for-loop and it seems to be working fine now.

aqeelat:
@hramezani I updated the PR with the comments you suggested.

hramezani:
Thanks @aqeelat LGTM :+1: 

Now, you need to write down the required steps for migration (as a comment to this PR). Then I will create a PR out of them later.

aqeelat:
Version 3.0.0 introduces breaking changes. Please review the migration guide below before upgrading.
If you're new to django-auditlog, you can ignore this part.

The major change in the version is that we're finally storing changes as json instead of json-text. To convert the existing records, this version has a database migration that does just that. However, this migration will take a long time if you have a huge amount of records, causing your database and application to be out of sync until the migration is complete.

To avoid this, follow these steps:
1. Before upgrading the package, add these two variables to `settings.py`:
  * `AUDITLOG_TWO_STEP_MIGRATION = True`
  * `AUDITLOG_USE_TEXT_CHANGES_IF_JSON_IS_NOT_PRESENT = True`
2. Upgrade the package. Your app will now start storing new records as JSON, but the old records will accessible via `LogEntry.changes_text`.
3. Use the newly added `auditlogjsonmigration` command to migrate your records. Run `django-admin auditlogjsonmigration --help` to get more information.
4. Once all records are migrated, remove the variables listed above, or set their values to False.

aqeelat:
Feel free to reword this as you see fit.
@hramezani

hramezani:
Thanks @aqeelat 

I've created issue https://github.com/jazzband/django-auditlog/issues/503 for this.

I am going to test the management command in my machine and will merge it if everything works fine.
This will happen probably next week

aqeelat:
Any updates here? @hramezani

hramezani:
@aqeelat Not yet!

I am too busy nowadays and as this is required for V3, it's not very urgent for me.
I also need to add documentation for this.

aqeelat:
Can you at least release the cid bits as a 2.x version?Sent from my iPhoneOn Apr 14, 2023, at 11:36 AM, Hasan Ramezani ***@***.***> wrote:﻿
@aqeelat Not yet!
I am too busy nowadays and as this is required for V3, it's not very urgent for me.
I also need to add documentation for this.

hramezani:
@aqeelat I am ok to release it in V2, we need to:

- Cherry-pick [cid commit](https://github.com/jazzband/django-auditlog/commit/bc6d393390e5f746df18b07c5dfd82d6fac09c74) on top of [V2](https://github.com/jazzband/django-auditlog/tree/v2)
-  Make a new release for V2(V2.3)

Would you like to work on it?

aqeelat:
@hramezani When would you be able to approve the PR?

hramezani:
> @hramezani When would you be able to approve the PR?

@aqeelat  we need to document how to do the migration. You already provided the required step for that and we have an [issue](https://github.com/jazzband/django-auditlog/issues/503).
My initial plan was to work on documentation for this PR but I am too busy :( 
It would be great if you add the documentation if you have time, them we are good to go

aqeelat:
@hramezani I'll get something out this weekend ISA.

hramezani:
> @hramezani I'll get something out this weekend ISA.

Great! Please add the documentation to this PR

aqeelat:
@hramezani pushed

hramezani (dismissed):

auditlog_tests/test_two_step_json_migration.py
```diff
@@ -0,0 +1,151 @@
...
+    def test_native_unsupported(self):
+        # Arrange
+        log_entry = self.make_logentry()
+
+        # Act
+        outbuf, errbuf = self.call_command("-d=oracle")
+        log_entry.refresh_from_db()
+
+        # Assert
+        msg = "Not yet implemented. Run this management commad without passing a -d/--database argument."
```
hramezani, line 134:
```suggestion
        msg = "Not yet implemented. Run this management command without passing a -d/--database argument."
```

hramezani, line 134:
```suggestion
        msg = "Not yet implemented. Run this management command without passing a -d/--database argument."
```

hramezani (approved):

Issue #493, opened 2023-01-04
Can't remove any auditlogged objects through the admin


aleh-rymasheuski:
As @madomdy [commented](https://github.com/jazzband/django-auditlog/pull/449#issuecomment-1370397139) in a related pull request, they are unable to remove objects of any model connected to the audit log.

![screenshot](https://user-images.githubusercontent.com/5181983/210469079-1eda4086-90cb-4334-a937-112d513234c4.png)

Possibly minimal app to reproduce the issue (I haven't tested it):
```python
# model.py
from auditlog.models import AuditlogHistoryField
from django.db import models


class MyModel(models.Model):
    history = AuditlogHistoryField()    

# admin.py
from django.contrib import admin

from .models import MyModel


admin.site.register(MyModel)
```

Instances of `MyModel` can't be removed through the admin because of `AuditlogHistoryField` and the delete permissions retracted in #449.

Workaround: pass `delete_related=False` to `AuditlogHistoryField` (`history = AuditlogHistoryField(delete_related=False)`).

hramezani:
I also couldn't reproduce the problem

ilhan:
Seems to be related to AuditlogHistoryField, I added `delete_related=False` to AuditlogHistoryField(delete_related=False) as a fix, haven't investigated further.

alieh-rymasheuski:
Thanks @ilhan, that was the missing piece of the puzzle. I'll update the description of the issue.

christianwgd:
Wouldn't it be much more flexible, if we can not only set this to True or False but have the same values like in normal referential integrity (SET_NULL, CASCADE, PROTECT)?

alieh-rymasheuski:
@christianwgd, your comment made me investigate this flag more deeply. I had the impression that `delete_related` was a standard parameter of `GenericRelation` and I was about to say so. I was wrong.

At the first glance it makes sense to support `on_delete` keyword. But on the other hand, how valuable `SET_NULL` (and the rest of the family of `SET` options) may be for the audit log table?

@hramezani, if we change `delete_related` to be `False` by default (or make an equivalent change through the potential new `on_delete` parameter), this will be a breaking change. We should include it in the 3.0 release.

christianwgd:
[...]

PR #496, opened 2023-01-16
Revert #449 "Disallow changing or deleting log entries"

alieh-rymasheuski:
As described in #493, this change inadvertently breaks deletion of auditlogged objects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Next Release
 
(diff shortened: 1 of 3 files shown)


alieh-rymasheuski:
@hramezani, please make this a patch release 2.2.2 because #449 was part of a patch release 2.2.1.

hramezani (approved):

Issue #467, opened 2022-12-09
Feature Request: Global Exclude for Field Names?


darkpixel:
I would love to see a feature where you could globally exclude field names.

I want all my of models audited using ```AUDITLOG_INCLUDE_ALL_MODELS = True``` because there are hundreds of them, and I don't want to configure them all by hand (potentially forgetting in the future as models are added).

...but every model I have inherits from a 'base model' that looks like this:

```
class BaseModel(models.Model):                                                                                              
                                                                                                                         
    """ BaseModel provides information on creation and modification times for all child objects. """                     
                                                                                                                         
    created = models.DateTimeField(editable=False, blank=True, null=True)                                                
    modified = models.DateTimeField(editable=False, blank=True, null=True)                                               
    created_by = models.ForeignKey(get_user_model(), null=True, blank=True, on_delete=models.SET_NULL, related_name='%(app_label)s_%(class)s_created_by')
    modified_by = models.ForeignKey(get_user_model(), null=True, blank=True, on_delete=models.SET_NULL, related_name='%(app_label)s_%(class)s_modified_by')
                                                                                                                         
    def save(self, *args, **kwargs):                                                                                     
        """ Save an object and note the created or modified time. """                                                    
        if not self.pk:                                                                                                  
            self.created = timezone.now()                                                                                
        else:                                                                                                            
            self.modified = timezone.now()                                                                               
        super(BaseModel, self).save(*args, **kwargs)                                                                     
                                                                                                                         
    class Meta(object):                                                                                                  
        abstract = True   
```

I would love a global setting that would prevent logging changes to those fields on every model.

i.e. ```AUDITLOG_EXCLUDE_FIELDS = ['created', 'created_by', 'modified', 'modified_by']```

The use-case is that I have a tool that updates models over an API and is constantly updating tens of thousands of them every few minutes.  Most of the time the data on the model doesn't change.  i.e. the "name" field is set to Joe and the tool posts {"name": "joe"}...so the only change to the model is the 'modified' and 'modified_by' fields.  I'd like to easily suppress that in the logs.

hramezani:
Thanks @darkpixel for this issue.

Good idea we can add the `AUDITLOG_EXCLUDE_FIELDS` setting to exclude fields.

Important note: This setting has to be considered when `AUDITLOG_INCLUDE_ALL_MODELS=True`. we have same conditions for `AUDITLOG_EXCLUDE_TRACKING_MODELS`

hramezani:
[...]

PR #498, opened 2023-01-18
Add ability to globally exclude fields by name on all models

darkpixel:
My first stab at solving #467.

Adds the ability to list field names in ```AUDITLOG_EXCLUDE_TRACKING_MODELS``` that will be excluded from logging.
If a field in ```AUDITLOG_EXCLUDE_TRACKING_MODELS``` is the only field changed, nothing will be logged.

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -189,6 +189,23 @@ You can use this setting to register all your models:
 
 .. versionadded:: 2.1.0
 
+**AUDITLOG_EXCLUDE_TRACKING_FIELDS**
+
+You can use this setting to exclude named fields from ALL models.
+This is useful when lots of models share similar fields like
+```created``` and ```modified``` and you want those excluded from
+logging.
+It will be considered when ``AUDITLOG_INCLUDE_ALL_MODELS`` is `True`.
+
+.. code-block:: python
+
+    AUDITLOG_EXCLUDE_TRACKING_FIELDS = (
+        "created",
+        "modified"
+    )
+
+.. versionadded:: 2.2.3
+
 **AUDITLOG_EXCLUDE_TRACKING_MODELS**
 
 You can use this setting to exclude models in registration process.
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -204,7 +204,7 @@ It will be considered when ``AUDITLOG_INCLUDE_ALL_MODELS`` is `True`.
         "modified"
     )
 
-.. versionadded:: 2.2.3
+.. versionadded:: 3.0.0
 
 **AUDITLOG_EXCLUDE_TRACKING_MODELS**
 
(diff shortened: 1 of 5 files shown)


hramezani:
Thanks @darkpixel for this patch.

You need to consider:

- Add a check for the value of this setting to be `list` or `tuple` like https://github.com/jazzband/django-auditlog/blob/f6b9e9f931ca663c8b08c9afd6b59773d1c734da/auditlog/registry.py#L282-L285
- Add a check to only enable this setting with `AUDITLOG_INCLUDE_ALL_MODELS` like https://github.com/jazzband/django-auditlog/blob/f6b9e9f931ca663c8b08c9afd6b59773d1c734da/auditlog/registry.py#L287-L293
- Add some test for this change 
- Add a changelog entry in https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md

docs/source/usage.rst
```diff
@@ -189,6 +189,23 @@ You can use this setting to register all your models:
...
+It will be considered when ``AUDITLOG_INCLUDE_ALL_MODELS`` is `True`.
+
+.. code-block:: python
+
+    AUDITLOG_EXCLUDE_TRACKING_FIELDS = (
+        "created",
+        "modified"
+    )
+
+.. versionadded:: 2.2.3
```
hramezani, line 207:
```suggestion
.. versionadded:: 3.0.0
```

darkpixel:
Thanks @hramezani 

I should be able to get those changed pushed later today!

hramezani:
Great @darkpixel 
Please rebase your patch with `master` because the failing CI in Python 3.8 and 3.9 got fixed in https://github.com/jazzband/django-auditlog/commit/7a7e2eb2efba3739ddb202c5e2fcfcfda7ec86e1

darkpixel:
Something's not making sense to me with the test framework.

I verified that the changes I've made actually work in my production application--fields listed in ```AUDITLOG_EXCLUDE_TRACKING_FIELDS = ('created', 'modified')``` prevents changes to those fields from being logged.

But given this test case, they are being logged:
```
class GlobalExcludeFieldTest(TestCase):
    """ Do not log changes to globally excluded fields. """

    def setUp(self):
        self.obj = SimpleModel.objects.create(text='Testing')

    def test_update_excluded_field(self):
        """ Test that updates to excluded fields are not logged """
        with override_settings(AUDITLOG_INCLUDE_ALL_MODELS=True) and override_settings(AUDITLOG_EXCLUDE_TRACKING_FIELDS=('datetime',)):
                obj = self.obj
                obj.datetime = datetime.datetime.now()
                obj.save()
                                                                                                                                                                                                                                                
                self.assertEqual(
                    obj.history.filter(action=LogEntry.Action.UPDATE).count(),
                    0,
                    msg="There are no log entries",
                )
 ```

I've poked through all the tests, and it seems like it should be working...but my guess is ```auditlog``` is already initialized with the wrong settings by the time my test starts?

hramezani:
@darkpixel Thanks for your update.

Your branch is protected and I am not allowed to push to your PR. Could you please allow me? I've fixed the problem and want to push.

darkpixel:
> Could you please allow me? I've fixed the problem and want to push.

Added!

hramezani:
@darkpixel Thanks!

Please check my changes and if it is ok for you, I will merge it

darkpixel:
```self.test_auditlog.register_from_settings()```

Aah--that's what I was missing.

Do I need to go into any more depth on the tests?  i.e. create a model, change an excluded field, and then check to make sure it wasn't saved as part of the changes dict?

hramezani:
No, I think it is enough to test it at this level. no need to go deeper

hramezani (approved):

Issue #502, opened 2023-01-25
Missing migration in master HEAD?


darkpixel:
I think there's a missing migration in HEAD on the master branch.

I added HEAD to my project, nuked the ```auditlog_logentry``` table and nuked the migrations (```delete from django_migrations where app = 'auditlog';```) and then ran ```python manage.py migrate```.

After that if I run ```python manage.py makemigrations``` I get prompted that ```It is impossible to change a nullable field 'cid' on logentry to non-nullable without providing a default. This is because the database needs something to populate existing rows.```

darkpixel:
The down-side to this is that if I hit option 3 to "Quit and manually define a default value in models.py", it won't make migrations for any of my other apps.

On a side note, I think there's a bug in Django because ```python manage.py makemigrations myapp``` also tries to make migrations for django-auditlog, otherwise this would basically be a non-issue.

Anyways, when I choose option 2, it does create a migration for the cid field:

```
It is impossible to change a nullable field 'cid' on logentry to non-nullable without providing a default. This is because the database needs something to populate existing rows.
Please select a fix:
 1) Provide a one-off default now (will be set on all existing rows with a null value for this column)
 2) Ignore for now. Existing rows that contain NULL values will have to be handled manually, for example with a RunPython or RunSQL operation.
 3) Quit and manually define a default value in models.py.
Select an option: 2
Migrations for 'auditlog':
  /home/aaron/.virtualenvs/-redacted-/lib/python3.9/site-packages/auditlog/migrations/0016_alter_logentry_cid.py
    - Alter field cid on logentry
```

The migration it creates looks like this:
```
# Generated by Django 4.1.4 on 2023-01-25 16:20    
    
from django.db import migrations, models    
    
    
class Migration(migrations.Migration):    
    
    dependencies = [    
        ("auditlog", "0015_alter_logentry_changes"),    
    ]    
    
    operations = [    
        migrations.AlterField(    
            model_name="logentry",    
            name="cid",    
            field=models.CharField(    
                blank=True, db_index=True, max_length=255, verbose_name="Correlation ID"    
            ),    
        ),    
    ]   
```

darkpixel:
Additionally, applying that generated migration leads to errors:
```django.db.utils.IntegrityError: null value in column "cid" violates not-null constraint```

hramezani:
Thanks @darkpixel for reporting.

Yes, you are right. I think we can fix the problem by adding `null=True` to cid.

```diff
[...]

PR #506, opened 2023-02-03
Add null=True to ci field

hramezani:
fixes https://github.com/jazzband/django-auditlog/issues/502


hramezani:
@darkpixel could you please take a look?

aqeelat (approved):
A major case of “oops, my bad.” 🌚 sorry guys.

PR #508, opened 2023-02-07
fixed getting field's verbose_name

AllYouZombies:
Sometimes, for example, when using a custom db router in microservices, there may not be direct access to the application / model. In such cases, we need te get at least field name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - fix: Fix a bug in audit log admin page when `USE_TZ=False`. ([#511](https://github.com/jazzband/django-auditlog/pull/511))
 - fix: Make sure `LogEntry.changes_dict()` returns an empty dict instead of `None` when `json.loads()` returns `None`. ([#472](https://github.com/jazzband/django-auditlog/pull/472))
 - fix: Always set remote_addr even if the request has no authenticated user. ([#484](https://github.com/jazzband/django-auditlog/pull/484))
+- fix: Fix a bug in getting field's `verbose_name` when model is not accessible. ([508](https://github.com/jazzband/django-auditlog/pull/508))
 
 ## 2.2.2 (2023-01-16)
 
--- a/auditlog/mixins.py
+++ b/auditlog/mixins.py
@@ -139,7 +139,7 @@ class LogEntryAdminMixin:
 
     def field_verbose_name(self, obj, field_name: str):
         model = obj.content_type.model_class()
-        if not model:
+        if model is None:
             return field_name
         try:
             model_fields = auditlog.get_model_fields(model._meta.model)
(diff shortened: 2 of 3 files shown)


HarleyK1ng:
It may not be easy to reproduce the issue, since not many people build microservices with Django using DB routers.
Pretend that there is content_type in db, but no app containing model

hramezani:
Thanks @HarleyK1ng for this patch.

Please try to include a test for this patch and if it's hard for you please create a minimal project to reproduce the bug. then I might help you. otherwise, we can try to add a test by mocking.

Please add a changelog entry as well.

HarleyK1ng:
Hi!

Yes, creating a tests is actually hard for me.
So I created minimal sample project to show in which cases issue appears.
Please, see this: https://github.com/HarleyK1ng/auditlog_minimal
It only requires **docker compose** to run project.

There's _3 containers_ in **docker compose**:
 - Postgres DB to share auth data between apps
 - parent app that provides authentication
 - child app that uses shared authentication

Child app uses DB router, that points to _shared auth DB_ if auth data is requested.
_Auditlog's_ data should be stored in shared DB too, since it _has relations to auth models_.
In addition storing logs in shared db allows to see them from 1 place.

Since parent app does not have child app's models, it can't retrieve `verbose_name` from child app's models.

Because of this I added "`if`" statement to check if model exists or data comes from different Django project.

hramezani:
@HarleyK1ng We can add a test by patching. something like:


```diff
diff --git a/auditlog_tests/tests.py b/auditlog_tests/tests.py
index 487a45d..ec74de5 100644
--- a/auditlog_tests/tests.py
+++ b/auditlog_tests/tests.py
@@ -16,6 +16,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AnonymousUser, User
 from django.contrib.contenttypes.models import ContentType
 from django.core import management
+from django.db.models import ForeignKey
 from django.db.models.signals import pre_save
 from django.test import RequestFactory, TestCase, override_settings
 from django.urls import reverse
@@ -1625,6 +1626,20 @@ class DiffMsgTest(TestCase):
         # Re-register
         auditlog.register(SimpleModel)
 
+    def test_field_verbose_name(self):
+        log_entry = self._create_log_entry(
+            LogEntry.Action.CREATE,
+            {"test": "test"},
+        )
+
+        self.assertEqual(self.admin.field_verbose_name(log_entry, "actor"), "Actor")
+
+        with patch(
+            "django.contrib.contenttypes.models.ContentType.model_class",
+            return_value=None,
+        ):
+            self.assertEqual(self.admin.field_verbose_name(log_entry, "actor"), "actor")
+
 
 class NoDeleteHistoryTest(TestCase):
     def test_delete_related(self):
```

Also, Please add a changlog entry for this bugfix in https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md

hramezani (approved):

PR #487, opened 2022-12-28
Fix localtime bug

ZahraEbrahimi01:
fix #480 bug

PR #511, opened 2023-02-10
Fix a bug in audit log admin page when `USE_TZ=False`

hramezani:
Continue to https://github.com/jazzband/django-auditlog/pull/487

Thanks @nbensa for providing the test case


aqeelat (approved):
LGTM

PR #512, opened 2023-02-10
Remove first_name / last_name from admin if missing from model

### LINKED ISSUE #509: Document or remove first_name / last_name as a search in the admin
If you use a custom user that does not include `first_name` / `last_name` as a field, searching in the admin will error out as `actor__first_name` and `actor__last_name` are not valid fields.

I can think of a few options:

1.  Unregister the admin using `admin.site.unregister(LogEntry)`
2. Subclass `LogEntryAdmin`, override `search_fields`
3. Register this new admin.

Or simply remove these fields from the built-in admin `search_fields`.

Or only add these search is `get_user_model` + meta ensures they exist would help.

I generally avoid splitting up first name / last name these days, though I know Django defaults to this.

Let me know which option sounds reasonable and I'm happy to open a PR. Great project, saved me a lot of time. Thanks.

silviogutierrez:
Closes #509 

This is a "static" approach, which I think is better than overriding `get_search_fields`. I believe the `get_*` methods are for truly dynamic behavior (based on the request, the current user, etc). With this, admin checks will still apply (though not particularly useful).

### DIFF
--- a/auditlog/admin.py
+++ b/auditlog/admin.py
@@ -31,7 +31,11 @@ class LogEntryAdmin(admin.ModelAdmin, LogEntryAdminMixin):
         "object_repr",
         "changes",
         f"actor__{user_model.USERNAME_FIELD}",
-    ] + (["actor__first_name", "actor__last_name"] if has_first_and_last_name_fields else [])
+    ] + (
+        ["actor__first_name", "actor__last_name"]
+        if has_first_and_last_name_fields
+        else []
+    )
     list_filter = ["action", ResourceTypeFilter, CIDFilter]
     readonly_fields = ["created", "resource_url", "action", "user_url", "msg"]
     fieldsets = [
--- a/auditlog/admin.py
+++ b/auditlog/admin.py
@@ -1,7 +1,8 @@
 from django.contrib import admin
 from django.contrib.auth import get_user_model
+from django.utils.translation import gettext_lazy as _
 
-from auditlog.filters import ResourceTypeFilter
+from auditlog.filters import CIDFilter, ResourceTypeFilter
 from auditlog.mixins import LogEntryAdminMixin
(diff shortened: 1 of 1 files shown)


hramezani:
@silviogutierrez Thanks for this path.

As I mention in the issue, overriding the `get_search_fields` would be a better approach. please implement it like this.
Also, you need to add some tests and a changelog entry in https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md

auditlog/admin.py
```diff
@@ -22,10 +30,12 @@ class LogEntryAdmin(admin.ModelAdmin, LogEntryAdminMixin):
         "timestamp",
         "object_repr",
         "changes",
-        "actor__first_name",
-        "actor__last_name",
-        f"actor__{get_user_model().USERNAME_FIELD}",
-    ]
+        f"actor__{user_model.USERNAME_FIELD}",
+    ] + (
```
RomHartmann, line 34: Something like the following adds whichever field is available, handling the case where either `fist_name` or `last_name` do actually exist

```
    search_fields = [
        "timestamp",
        "object_repr",
        "changes",
        f"actor__{get_user_model().USERNAME_FIELD}",
    ] + [f"actor__{f.name}" for f in get_user_model()._meta.get_fields() if f.name in ["first_name", "last_name"]]
```

hramezani: This is better because includes whatever fields that are available.
Feel free to create a new PR with tests and changelog

hramezani:
The approach of this PR should be changed.
Also test and change log has to be included.

BTW, you can create your custom admin inherited from `LogEntryAdmin` and implement your desired `get_search_fields` to unblock yourself or even better create a new PR and fix it for the library

PR #520, opened 2023-03-31
Fix GitHub actions deprecations

alieh-rymasheuski:
See these announcements:
- https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


hramezani (approved):
Thanks!

PR #525, opened 2023-05-05
Audit changes to FK fields when saved using *_id naming.

barqshasbite:
In Django you can save changes to a FK field on a model using the ID field name (attname) in addition to the regular FK field name with:

    model.related_model_id = 42
    model.save(update_fields=["related_model_id'])

or:

    model.related_model = related_model
    model.save(update_fields=["related_model_id'])

as opposed to the more common:

    model.related_model = related_model
    model.save(update_fields=["related_model'])

This change ensures those model changes are logged properly.

### DIFF
--- a/auditlog/diff.py
+++ b/auditlog/diff.py
@@ -147,14 +147,14 @@ def model_instance_diff(
         model_fields = None
 
     if fields_to_check:
-        _fields = set()
-        for field in fields:
-            if isinstance(field, ForeignKey):
-                if field.attname in fields_to_check:
-                    _fields.add(field)
-            if field.name in fields_to_check:
-                _fields.add(field)
-        fields = _fields
+        fields = {
+            field
+            for field in fields
+            if (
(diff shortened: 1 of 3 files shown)


auditlog/diff.py
```diff
@@ -147,7 +147,14 @@ def model_instance_diff(
         model_fields = None
 
     if fields_to_check:
-        fields = {field for field in fields if field.name in fields_to_check}
+        _fields = set()
```
hramezani, line 150:
```suggestion
        fields = {
            field
            for field in fields
            if (
                (isinstance(field, ForeignKey) and field.attname in fields_to_check)
                or (field.name in fields_to_check)
            )
        }

```

hramezani:
Thanks @barqshasbite for refreshing the old PR 👍 
I've left a suggestion
Please add a changelog as well

hramezani (approved):

PR #540, opened 2023-07-03
Drop Django 4 support

hramezani:


aleh-rymasheuski (approved):

hramezani:
FYI, it just removed Django 4.0 support but 4.1 and 4.2 still support.


BTW, it's not yet deployed and will be deployed in the next release probably in V3

aleh-rymasheuski:
This change should be read that V3 will no longer _officially_ support Django 4.

Issue #526, opened 2023-05-07
Serializing fields with F expressions doesn't work


mcanu:
I'm using `django-auditlog` to track the changes in a model `Pack` that has a `quantity` field. To avoid potential race conditions I update that field with [Django's F expressions](https://docs.djangoproject.com/en/4.2/ref/models/expressions/#f-expressions). However, when I save the model instance, I get a `ValidationError` with `“F(quantity) + Value(1)” value must be an integer.`.

Example:
```python
from auditlog.registry import auditlog
from django.db import models

class Pack:
    quantity = models.PositiveSmallIntegerField()

auditlog.register(Pack, serialize_data=True)

pack = Pack.objects.first()
pack.quantity = models.F('quantity') + 1
pack.save()

Exception: django.core.exceptions.ValidationError: ['“F(quantity) + Value(1)” value must be an integer.']
```
This only happens when `serialize_data=True`. Is there a way to work around this issue?

hramezani:
Thanks @mcanu for reporting.

@sum-rock Could you please take a look?

hramezani:
I've investigated the problem and the error comes from the below line in `_get_copy_with_python_typed_fields`

https://github.com/jazzband/django-auditlog/blob/875c6b789f1dccacfd8c0568e050768fe5572636/auditlog/models.py#L276

So, I think we can add `try ... except` on this line and prevent adding fields that have a problem in `serialized_data`.

There might be a better way to handle it. @sum-rock do you have any idea?

PR #544, opened 2023-07-12
Fix a bug in `serialized_data` with F expressions

hramezani:
Closes https://github.com/jazzband/django-auditlog/issues/526


peymanslh (approved):
LGTM

PR #556, opened 2023-08-24
Allow cascade deletion of auditlog entries

### LINKED ISSUE #493: Can't remove any auditlogged objects through the admin

aleh-rymasheuski:
Resolves #493.

This pull request allows cascade deletion of log entries, but still disallows direct deletion through the admin.

### DIFF
--- a/auditlog/admin.py
+++ b/auditlog/admin.py
@@ -40,6 +40,11 @@ class LogEntryAdmin(admin.ModelAdmin, LogEntryAdminMixin):
         return False
 
     def has_delete_permission(self, request, obj=None):
+        if request.resolver_match and request.resolver_match.url_name not in [
+            pattern.name for pattern in self.urls if pattern.name
+        ]:
+            # only allow cascade delete to satisfy delete_related flag
+            return super().has_delete_permission(request, obj)
         return False
 
     def get_queryset(self, request):
--- a/auditlog/admin.py
+++ b/auditlog/admin.py
@@ -1,3 +1,5 @@
+from functools import cached_property
+
 from django.contrib import admin
 from django.contrib.auth import get_user_model
 from django.utils.translation import gettext_lazy as _
@@ -39,10 +41,15 @@ class LogEntryAdmin(admin.ModelAdmin, LogEntryAdminMixin):
     def has_change_permission(self, request, obj=None):
         return False
 
+    @cached_property
+    def _own_url_names(self):
+        return [pattern.name for pattern in self.urls if pattern.name]
+
     def has_delete_permission(self, request, obj=None):
-        if request.resolver_match and request.resolver_match.url_name not in [
-            pattern.name for pattern in self.urls if pattern.name
-        ]:
+        if (
+            request.resolver_match
+            and request.resolver_match.url_name not in self._own_url_names
+        ):
             # only allow cascade delete to satisfy delete_related flag
             return super().has_delete_permission(request, obj)
         return False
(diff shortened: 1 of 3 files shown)


aleh-rymasheuski:
@hramezani, we don't have a structure for testing this case with automated tests, and I don't have an elegant way to test this. If you can think of a good way of testing this, please add a test. Otherwise I'd suggest merging this change without a test.

auditlog/admin.py
```diff
@@ -40,6 +40,11 @@ def has_change_permission(self, request, obj=None):
         return False
 
     def has_delete_permission(self, request, obj=None):
+        if request.resolver_match.url_name not in [
+            pattern.name for pattern in self.urls if pattern.name
+        ]:
```
hramezani, line 45: @aleh-rymasheuski Could you please explain more about what is happening here?

Also, I think you can move the `[pattern.name for pattern in self.urls if pattern.name]` out of method and define a class attr to don't repeat the loop in each call.

hramezani:
> @hramezani, we don't have a structure for testing this case with automated tests, and I don't have an elegant way to test this. If you can think of a good way of testing this, please add a test. Otherwise I'd suggest merging this change without a test.

Thanks @aleh-rymasheuski for this patch 🙏 

I think we can test it like:

```diff
diff --git a/auditlog_tests/tests.py b/auditlog_tests/tests.py
index 12a4725..8824780 100644
--- a/auditlog_tests/tests.py
+++ b/auditlog_tests/tests.py
@@ -19,7 +19,7 @@ from django.core import management
 from django.db import models
 from django.db.models.signals import pre_save
 from django.test import RequestFactory, TestCase, override_settings
-from django.urls import reverse
+from django.urls import resolve, reverse
 from django.utils import dateformat, formats
 from django.utils import timezone as django_timezone
 from django.utils.encoding import smart_str
@@ -1555,6 +1555,9 @@ class DiffMsgTest(TestCase):
         super().setUp()
         self.site = AdminSite()
         self.admin = LogEntryAdmin(LogEntry, self.site)
+        self.user = User.objects.create_user(
+            username="test", email="test@example.com", password="top_secret"
+        )
 
     def _create_log_entry(self, action, changes):
         return LogEntry.objects.log_create(
@@ -1711,6 +1714,13 @@ class DiffMsgTest(TestCase):
         ):
             self.assertEqual(self.admin.field_verbose_name(log_entry, "actor"), "actor")
 
+    def test_has_delete_permission(self):
+        request = RequestFactory().get("/simplemodel/1/")
+        request.resolver_match = resolve("/simplemodel/1/")
+        request.user = self.user
+
+        self.assertFalse(self.admin.has_delete_permission(request))
+
 
 class NoDeleteHistoryTest(TestCase):
     def test_delete_related(self):
```

aleh-rymasheuski:
@hramezani, I added a test based on your suggestion.

auditlog/admin.py
```diff
@@ -40,6 +40,11 @@ def has_change_permission(self, request, obj=None):
         return False
 
     def has_delete_permission(self, request, obj=None):
+        if request.resolver_match and request.resolver_match.url_name not in [
+            pattern.name for pattern in self.urls if pattern.name
```
hramezani, line 44: can't we move the admin urls patterns list creation out of this method to not calculate it every time?

aleh-rymasheuski: I'm not sure when in the lifecycle of the admin `self.urls` is available, so I'll just move this expression into a cached property.

aleh-rymasheuski: Done.

hramezani (approved):

PR #557, opened 2023-08-24
Set history delete_related to False by default

aleh-rymasheuski:
Based on the discussion in #493 and my comment at https://github.com/jazzband/django-auditlog/issues/493#issuecomment-1383913950, we should not delete audit log entries together with the tracked object by default.


hramezani (approved):

hramezani:
Thanks @aleh-rymasheuski

PR #558, opened 2023-08-30
Pass instance_old to post_log.send

ilius:
We need to send audit log to logstash (this can apply to any external database/API), so we need the old model object to extract the changes.


hramezani (approved):

hramezani:
Thanks @ilius.

It will be release in `V3.0.0`.

PR #559, opened 2023-08-31
Stop deleting log entries in log_create

aleh-rymasheuski:
This pull request addresses two issues:
1. Even if a pk value is reused, we should not delete log entries for old deleted object. Old object's lifetime is easily identifiable by a pair of matching created & deleted log entries.
2. (why I started to look into this in the first place) `log_create` spends 22% of its run time calling `delete()` method in a test run of a medium-sized data loading job on my machine (7% of the total run time of the job).

I assume this is technically a breaking change so I'd like it included in 3.0.0 release too.

Notice the change in the `log_update` handler. Before my changes, every object creation was producing two log entries: `updated` log entry, then `created` log entry; and while creating this second log entry, auditlog was deleting the first one, so it was not noticeable.

### DIFF
--- a/auditlog/models.py
+++ b/auditlog/models.py
@@ -66,26 +66,6 @@ class LogEntryManager(models.Manager):
             if callable(get_additional_data):
                 kwargs.setdefault("additional_data", get_additional_data())
 
-            # Delete log entries with the same pk as a newly created model.
-            # This should only be necessary when a pk is used twice.
-            if kwargs.get("action", None) is LogEntry.Action.CREATE:
-                if (
-                    kwargs.get("object_id", None) is not None
-                    and self.filter(
-                        content_type=kwargs.get("content_type"),
-                        object_id=kwargs.get("object_id"),
-                    ).exists()
-                ):
-                    self.filter(
-                        content_type=kwargs.get("content_type"),
-                        object_id=kwargs.get("object_id"),
-                    ).delete()
-                else:
-                    self.filter(
-                        content_type=kwargs.get("content_type"),
-                        object_pk=kwargs.get("object_pk", ""),
-                    ).delete()
-
             # set correlation id
             kwargs.setdefault("cid", get_cid())
             return self.create(**kwargs)
--- a/auditlog/receivers.py
+++ b/auditlog/receivers.py
@@ -49,7 +49,7 @@ def log_update(sender, instance, **kwargs):
 
     Direct use is discouraged, connect your model through :py:func:`auditlog.registry.register` instead.
     """
-    if instance.pk is not None:
+    if not instance._state.adding:
         update_fields = kwargs.get("update_fields", None)
         old = sender.objects.filter(pk=instance.pk).first()
         _create_log_entry(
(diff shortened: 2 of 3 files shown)


auditlog/receivers.py
```diff
@@ -52,6 +52,9 @@ def log_update(sender, instance, **kwargs):
     if instance.pk is not None:
         update_fields = kwargs.get("update_fields", None)
         old = sender.objects.filter(pk=instance.pk).first()
+        if not old:
```
aleh-rymasheuski, line 55: @hramezani, should I replace this check to `if instance._state.adding`? https://docs.djangoproject.com/en/4.2/ref/models/instances/#django.db.models.Model._state

aleh-rymasheuski: I thought about this for a short while and decided to use the `adding` flag (it exists for a reason).

hramezani (approved):

hramezani:
Thanks @aleh-rymasheuski 

@aqeelat it would be great to take a look at this PR if you have time.

auditlog/receivers.py
```diff
@@ -49,7 +49,7 @@ def log_update(sender, instance, **kwargs):
 
     Direct use is discouraged, connect your model through :py:func:`auditlog.registry.register` instead.
     """
-    if instance.pk is not None:
+    if not instance._state.adding:
```
aqeelat, line 52: good catch

auditlog/models.py
```diff
@@ -66,26 +66,6 @@ def log_create(self, instance, force_log: bool = False, **kwargs):
...
-                    self.filter(
-                        content_type=kwargs.get("content_type"),
-                        object_id=kwargs.get("object_id"),
-                    ).delete()
-                else:
-                    self.filter(
-                        content_type=kwargs.get("content_type"),
-                        object_pk=kwargs.get("object_pk", ""),
-                    ).delete()
-
```
aqeelat, line 88: Why was this added in the first place?
https://thoughtbot.com/blog/chestertons-fence

aqeelat: The primary key reuse is still a thing in mysql ([link](https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html#:~:text=In%20this%20case%20(when%20the,values%20normally%20are%20not%20reused.))). Should we handle this by checking the db engine? Or maybe by a configuration var?

aleh-rymasheuski: @aqeelat, I don't think we should *ever* delete auditlog entries by default. The audit log exists to be immutable.

aqeelat: @aleh-rymasheuski I agree. However, people who are serious about auditlogs will also be careful to avoid primary-key reuse.
The main issue is that when accessing the logs for a instance, we shouldn't get entries for older instances with the same object id.
I know this is an edge case, which is why I'm trying to find the most inclusive solution.

hramezani: Thanks @aqeelat for your input here. 

IMHO, we can merge it.

PR #561, opened 2023-09-06
fix: only fire the post_log signal when the log is created or when there is an error

### LINKED ISSUE #536: Custom signal post_log is triggered even when LogEntry is not created
From https://github.com/jazzband/django-auditlog/blob/master/auditlog/receivers.py#L99

```
    try:
        changes = model_instance_diff(
            diff_old, diff_new, fields_to_check=fields_to_check
        )

        if force_log or changes:
            LogEntry.objects.log_create(
                instance,
                action=action,
                changes=changes,
                force_log=force_log,
            )
    except BaseException as e:
        error = e
    finally:
        post_log.send(
            sender,
            instance=instance,
            action=action,
            error=error,
            pre_log_results=pre_log_results,
        )
        if error:
            raise error
```
When `changes` is None implying that the model was saved with no changes,  LogEntry object is not created, but the post_log signal is still being send in the `finally` block which shouldnt happen.

To reproduce the issue:
1. Set any registered model with a receiver for the post_log signal to print a log
2. Save an object without any changes

I am using the last master commit for this.

Another issue here is that the "Custom Signals" feature is not released yet but is part of the public documentation already - https://django-auditlog.readthedocs.io/en/latest/internals.html#custom-signals

aqeelat:
Closes #536

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Improvements
 
 #### Fixes
+* fix: only fire the `post_log` signal when the log is created or when there is an error in the process.
 
 ## 3.0.0-beta.1
 
--- a/auditlog/receivers.py
+++ b/auditlog/receivers.py
@@ -105,6 +105,8 @@ def _create_log_entry(
         action=action,
     )
     error = None
+    log_created = False
+    changes = None
     try:
         changes = model_instance_diff(
             diff_old, diff_new, fields_to_check=fields_to_check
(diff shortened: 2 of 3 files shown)


aqeelat:
@hramezani should we expose `force_log` or is that an internal variable?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 #### Improvements
 
 #### Fixes
+* only fire the `post_log` signal when the log is created or when there is an error in the process.
```
hramezani, line 10:
```suggestion
- fix: Only fire the `post_log` signal when the log is created or when there is an error in the process.
```

hramezani (dismissed):

hramezani:
Thanks @aqeelat for the path 🙏 

I only left a suggestion. Other than that, LGTM 👍 

> @hramezani should we expose force_log or is that an internal variable?

Not for now

aqeelat:
@hramezani then I guess we shouldn't pass it to the post_log signal?

hramezani:
> @hramezani then I guess we shouldn't pass it to the post_log signal?

yes, let's not pass it for now.

hramezani (approved):

hramezani:
Thanks @aqeelat

PR #571, opened 2023-10-03
fix: change verbose_name in changes_text migration

### LINKED ISSUE #570: [new user] Version 3.0.0 creating a migration file?
We've newly started using auditlog in our project and started directly with the 3.0.0 version.
Based on [this](https://github.com/jazzband/django-auditlog/issues/503) issue, we did not add the said flags as it says it can be ignored by new users

 Our project creates a migration file for verbose name change to `changes_text` field as the [models.py](https://github.com/jazzband/django-auditlog/blob/master/auditlog/models.py#L342) shows the verbose name
as "change message" but the [migrations](https://github.com/jazzband/django-auditlog/blob/master/auditlog/migrations/0015_alter_logentry_changes.py#L27) file has the verbose name as "text change message".

This looks like a mixup, I did try adding the flags to our project after which we stop seeing the new migration file but that does not feel like the right solution. From a new users perspective the models.py and the migrations file do not look like they are in sync.

aqeelat:
closes #570 
my bad.
I don't know how this happened. The good thing is that verbose_name doesn't alter the db, so we don't need to create a migration for people already on v3.

### DIFF
--- a/auditlog/migrations/0015_alter_logentry_changes.py
+++ b/auditlog/migrations/0015_alter_logentry_changes.py
@@ -24,7 +24,7 @@ def two_step_migrations() -> List:
         migrations.AddField(
             model_name="logentry",
             name="changes_text",
-            field=models.TextField(blank=True, verbose_name="text change message"),
+            field=models.TextField(blank=True, verbose_name="change message"),
         ),
         migrations.AlterField(
             model_name="logentry",


aqeelat:
I think the issue was because I created the migrations semi-manually (I was new to the project and didn't know how to set up a good local env)

hramezani:
V3 is in beta. So, people should not use it on prod. don't worry

hramezani:
We probably need to prepare the second beta for V3

hramezani (approved):

jgillard:
Was this meant to be included in 3.0b2? The 3.0b2 tag on Github is for an earlier commit from September, but there's a commit on master after this was merged that mentions preparing for 3.0b2, so is the tag wrong?

hramezani:
@jgillard seems you are right.

Not sure what was the problem. But I will prepare a new beta release soon

PR #572, opened 2023-10-03
Confirm Python 3.12 support

jgillard:
I followed the pattern set by the previous 3.11 MR: #447 

The tox envlist change might be debatable, but my understanding is that Python 3.12 will only be supported on 4.2 and above.

As my colleague mentioned in #570, we've just started using v3 of this package, so happy to chip in with small things like this. I thought it might want adding for 3.0. We know that we shouldn't be using v3 yet, but the risk level is fine for what we're doing.

Thanks for your efforts!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       max-parallel: 5
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
 
     services:
       postgres:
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @jgillard for this path 🙏 

Based on [Django supported Python versions](https://docs.djangoproject.com/en/4.2/faq/install/#what-python-version-can-i-use-with-django), Django not support Python 3.12.
[Django 5.0 supports Python 3.12](https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django) and will be released on `December 2023`.

So, let's keep `3.12` only for `main` branch for now.

jgillard:
You're correct about Django 4.2 not supporting Python 3.12 (although that may change at some point), so for now I've only applied it to django's main branch as you suggested.

hramezani (approved):

PR #581, opened 2023-11-02
use contextvar instead of threadlocal

hamedsh:
replace threading.local() with contextvars to be able to use with Django 4.2

### DIFF
--- a/auditlog/context.py
+++ b/auditlog/context.py
@@ -8,8 +8,8 @@ from django.db.models.signals import pre_save
 
 from auditlog.models import LogEntry
 
-auditlog_value = ContextVar('auditlog_value')
-auditlog_disabled = ContextVar('auditlog_disabled', default=False)
+auditlog_value = ContextVar("auditlog_value")
+auditlog_disabled = ContextVar("auditlog_disabled", default=False)
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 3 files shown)


auditlog/context.py
```diff
@@ -48,26 +48,30 @@ def _set_actor(user, sender, instance, signal_duid, **kwargs):
     This function becomes a valid signal receiver when it is curried with the actor and a dispatch id.
     """
     try:
-        auditlog = threadlocal.auditlog
-    except AttributeError:
+        auditlog = auditlog_value.get()
+    except LookupError as ex:
```
hramezani, line 52:
```suggestion
    except LookupError:
```

hramezani:
Thanks @hamedsh for this. I've added a commit to fix the lint error.

Let's see what are @aqeelat and @aleh-rymasheuski ideas here.

aleh-rymasheuski (commented):
Not familiar with the API so can't review the code properly, but this change seems logical.

My only concern is that the `threadlocal` name is public (though undocumented) API so we should (a) bump the "feature" part of the version number (or squeeze this change into the v3 major release) and (b) document this change in the changelog.

hramezani:
We will release it in V3.

Yeah, the change has to be mentioned in changelog

hramezani:
@hamedsh Could you please add an entry to changelog?

hramezani (approved):

Issue #586, opened 2023-11-12
Dynamic cross-instance register/unregister


kiblik:
Hello,

we are developing and running a Django app
- which is horizontally scalable on main Django instances as well as on Celery instances
- with a focus on allowing a user to configure as many parameters from inside of the application (without the necessity to change environment variables and/or restart instances)

In this setup, we would like to get users the ability to enable and disable audit-logging. 

*Question: Is this possible?* I do not think so. But I will be more than happy if somebody will provide instructions on how to achieve our goals.

The reason why it is not possible in my opinion: To enable/disable audit-logging for each model, it is necessary to call `auditlog.register(Model_Name)`/`auditlog.unregister(Model_Name)` and this call has only local effect. So change will happen only on the instance which processes the user's request and it won't be propagated to other Django and Celery instances.

Because of this, if we would like to stay scalable, we are forced to change the setting via environment variables before each instance is started.

Am I right or is there some way, to dynamically enable/disable audit-logging?

hramezani:
Please check [my comment](https://github.com/jazzband/django-auditlog/issues/589#issuecomment-1822749172) on a related issue. probably it would useful.

kiblik:
> Please check [my comment](https://github.com/jazzband/django-auditlog/issues/589#issuecomment-1822749172) on a related issue. probably it would be useful.

Thanks @hramezani. Yes, your proposal might be the way for our enviroment. Let's wait for implementation, review and release.

hramezani:
@kiblik We merged https://github.com/jazzband/django-auditlog/pull/590 and it will be available in the next beta release for testing.

kiblik:
@hramezani, looks nice, thank you.
I might be able to run a local test, but I cannot add to our project until your change is not part of an official (non-beta) release.

PR #590, opened 2023-11-23
Conditionally disable / enable logging

### LINKED ISSUE #589: Conditionally disable logging
Lets say I have an application where a user belongs to a company where audit logging is optional. 

Ideally I would like to attach a callback per model, that is run pre-save, and this callback should get the object to be saved and return a boolean if logging should be enabled / disabled.

Is there anyway I can do this today, by overriding some functionality? 

Thanks

#586 Looks like this is sort of a similar question

errrken:
If any receiver returns False, no logging will be made. This can be useful if logging should be conditionally enabled / disabled

Closes #589 possibly #586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next Release
 
-#### Breaking Changes
+#### Improvements
 - feat: If any receiver returns False, no logging will be made. This can be useful if logging should be conditionally enabled / disabled ([#590](https://github.com/jazzband/django-auditlog/pull/590))
 
 ## 3.0.0-beta.3 (2023-11-13)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next Release
 
+#### Breaking Changes
+- feat: If any receiver returns False, no logging will be made. This can be useful if logging should be conditionally enabled / disabled ([#590](https://github.com/jazzband/django-auditlog/pull/590))
+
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @errrken for this patch. please:

- Add some test for it in https://github.com/jazzband/django-auditlog/blob/c581a6e6476e1fe0afb9adfc252360e5d0987023/auditlog_tests/tests.py#L2340
- Add changelog entry in https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Next Release
 
+#### Breaking Changes
```
hramezani, line 5: I think it not a breaking change. as we are preparing django-auditlog for major release 3 and it's already in beta. 
```suggestion
#### Improvements
```

errrken:
Yeah I was just thinking if someone for some reason used to return False
before - it will suddenly disable their logging - that’s why I reasoned it
would be breaking…

fre 24 nov. 2023 kl. 11:57 skrev Hasan Ramezani ***@***.***>:

aleh-rymasheuski (approved):

errrken:
Thanks y'all, may I ask out of curiosity when we can expect it to be merged and when you expect 3.0 to be released?

hramezani (approved):

errrken:
> It will be released in the next beta release. I am not sure about v3 release time

Allright, so there will be one more beta? Any idea when that will be released?

Issue #454, opened 2022-11-17
Document how to pass `additional_data` 


hramezani:
Related issue https://github.com/jazzband/django-auditlog/issues/453
Example use case https://github.com/jazzband/django-auditlog/blob/2b0bc9efa2db94af2c5e130cce4e662cf3b2ad0e/auditlog_tests/models.py#L168-L178

PR #603, opened 2024-01-26
Add custom data and actor email

hamedsh:
### add actor_email
deleting a user from the Users table will make actor_id useless (its only a bunch of characters). by adding email as static content, it can be usable forever.

### adding custom_data 
most of the time, data change is not the only thing you need. for example in a real environment with a bunch of CLI actions and Crons, you need some additional records to specify the task that causes the change. by using custom_data, its possible to attach any data to the history log.
example:

```python
with set_auditlog_custom_data(custom_data={"type": "management_command", "command": "clear_unused_records")}):
    target_function(obj)
```

### DIFF
--- a/auditlog/admin.py
+++ b/auditlog/admin.py
@@ -29,9 +29,9 @@ class LogEntryAdmin(admin.ModelAdmin, LogEntryAdminMixin):
         f"actor__{get_user_model().USERNAME_FIELD}",
     ]
     list_filter = ["action", ResourceTypeFilter, CIDFilter]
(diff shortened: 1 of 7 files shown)


hramezani:
Thanks @hamedsh for this PR

The change for adding `email` filed is good.

The `LogEntry` model has an `additional_data` field and It can be used to provide additional data for the log record.
It has not been documented yet and we have an [issue](https://github.com/jazzband/django-auditlog/issues/454) for documenting this field. probably you can create a PR for that.

@hamedsh please close this PR and create a PR for the `actor email`.
If you want to make more changes please create separate PR for them.

PR #605, opened 2024-02-08
Include `LogEntry` instance in `post_log` signal

### LINKED ISSUE #576: post_save / post_log question
I have questions about `post_save` and `post_log`. In my requirement I want to take action on the object at the time it is being logged.

My first attempt was using `post_save`. Here's the issue: when I receive `post_save` for the `LogEntry`, at this moment, the object that is being logged is not saved yet. Is it possible to have the `post_save` of LogEntry _after_ the object was saved?

Second attempt was using the `post_log` signal instead. Here the object being logged is saved already, but another requirement is missing: The `post_log` signal does not include the actor. Is it possible to include it?

ppmathis:
Thank you for maintaining this great project! While working on a specific use case where I wanted to further process audit messages, I stumbled across the fact that the `post_log` signal which has been introduced in v3 somehow does not contain the actual instance of the `LogEntry` that has been created and saved.

This was rather confusing to me and makes it impossible to implement my use case, which also requires access to fields like `cid`, `remote_addr`, `timestamp` or `actor` - for the latter I even found an open issue #576 on this repository, where a user was also missing the possibility to access the actor of a `LogEntry`.

While all kinds of fields could be added to the signal, I hereby propose to simply pass the actual `LogEntry` instance as part of the signal when available, which I deem a simpler solution and avoids adding more and more arguments to the signal. This PR adds the `log_entry` argument, which is either an instance of `LogEntry` (= changes made/force-log and persisted) or `None` (= error occurred).

This will allow easily accessing all the available model fields in a signal or even directly interacting with the specific LogEntry instance in other ways. Merging this PR would fix #576 as well.

### DIFF
--- a/auditlog/receivers.py
+++ b/auditlog/receivers.py
@@ -113,7 +113,7 @@ def _create_log_entry(
         return
 
     error = None
-    log_created = False
+    log_entry = None
     changes = None
     try:
         changes = model_instance_diff(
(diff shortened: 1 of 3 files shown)


hramezani (approved):

hramezani:
Thanks @ppmathis

PR #609, opened 2024-02-13
Avoid exception in `changes_display_dict` when model is missing

ppmathis:
As django-auditlog v3 now stores the actual PKs for foreign models instead of their string representation, I switched my application from `changes_dict` to `changes_display_dict` so that the actual string-representation of foreign models is being shown again. While this worked in my local environment, it broke while trying out this change against a previously used testing database, specifically by raising a `KeyError`.

As it turns out, if a model has been previously registered with django-auditlog and made any entries, followed by then being no longer registered (e.g. if the model has been deprecated or further logging is no longer desired), any calls to `changes_display_dict` will fail. The reason behind this is that the method attempts to lookup the model within the registry, without checking if it actually still exists.

This lookup is only needed for respecting `mapping_fields` and has no other purpose. This PR extends the logic to first check if the model is contained in the registry, and only if that is the case, the actual lookup happens. If a previously existing model is being looked up instead, it will still show up properly, just without the mapped fields as they no longer exist.

### DIFF
--- a/auditlog/models.py
+++ b/auditlog/models.py
@@ -436,11 +436,14 @@ class LogEntry(models.Model):
         """
         :return: The changes recorded in this log entry intended for display to users as a dictionary object.
         """
-        # Get the model and model_fields
         from auditlog.registry import auditlog
 
+        # Get the model and model_fields, but gracefully handle the case where the model no longer exists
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks @ppmathis

PR #613, opened 2024-03-02
fix: error actor is none

morghim:


hramezani:
@morghim 

Could you please explain what problem are you going to solve?
Please share an example to show what is the problem

morghim:
> @morghim
> 
> Could you please explain what problem are you going to solve? Please share an example to show what is the problem

When I use library with rest framework any action user do save log with actor none , the library not catch user how make action 

What I do is use another library crum to catch user using thread local and I added to single make log

hramezani:
Have you checked the problem in `V3 Beta`? As you may know, we are working on `V3` major release. it is in Beta stage. You can install the `V3 beta` by `pip install django-auditlog --pre`. Please test the problem with the Beta first.

Preparing a small example case to reproduce the problem would be great. Please add a falling test.
Also, I don't want to add extra dependency to the library because of fixing this problem. please fix it with the existing functionality.

PR #623, opened 2024-04-02
Add new settings attr AUDITLOG_REGISTRY to allow custom auditlogs

### LINKED ISSUE #617: Have custom AuditLogModelRegistry
I'd like to have sth like this
```
from auditlog.registry import AuditlogModelRegistry

create_only_auditlog = AuditlogModelRegistry(
    create=True,
    update=False,
    delete=False,
    access=False,
    custom={
        post_save=my_signal_handler
    }
)

create_only_auditlog.register(MyModel)
```

But the problem is that we can not use this `create_only_auditlog`

because [model_instance_diff](https://github.com/jazzband/django-auditlog/blob/master/auditlog/diff.py#L108) would throw `KeyError` as it always get the default `auditlog`

Can we have new settings attribute to declare our custom auditlogs ?

hoangquochung1110:
This PR is to close #617

I've added a new configuration `AUDITLOG_REGISTRY`, paths to custom auditlog instances
which allow to configure models logging with flexible options (conditionally disabled/enabled logging and custom signal receivers)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## 3.0.0-beta.4 (2024-01-02)
 
 #### Improvements
+- feat: Add `AUDITLOG_REGISTRY` settings attribute to have multiple auditlogs with custom configurations ([#623](https://github.com/jazzband/django-auditlog/pull/623))
 - feat: Excluding ip address when `AUDITLOG_DISABLE_REMOTE_ADDR` is set to True ([#620](https://github.com/jazzband/django-auditlog/pull/620))
 - feat: If any receiver returns False, no logging will be made. This can be useful if logging should be conditionally enabled / disabled ([#590](https://github.com/jazzband/django-auditlog/pull/590))
 - Django: Confirm Django 5.0 support ([#598](https://github.com/jazzband/django-auditlog/pull/598))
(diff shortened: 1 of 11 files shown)


hramezani:
Thanks @hoangquochung1110 for the PR.

Honestly, the implementation looks complex to me. I was expecting a simple config that can be used instead of the default `auditlog` instance and I would suggest to have this for now.

BTW, if you are interested in implementing multiple auditlog instances at the same time, You can have a registry class to map every model to its auditlog instance. you can build the mapping in `register` method. then you can have multiple auditlog instance at the same time and the instances can be used for models registrations. something like:

```
from auditlog.registry import auditlog  # default one

custom_auditlog = AuditlogModelRegistry(create=False)

# here in `register` method we have to map the `MyModel` to default auditlog instance
auditlog.register(MyModel)

# here in `register` method we have to map the `MyModel1` to `custom_auditlog`
custom_auditlog.register(MyModel1)
```

Then in the other places of project, we have to find the auditlog instance based on model

hoangquochung1110:
Thanks @hramezani  for the review

I realize the implementation looks unnecessarily complex.

My initial goal is to disable signals and define custom signal handlers

I dived into the documentations
and the below snippet of code works for my case:

```
auditlog.register(
    Model,
)
pre_log.connect(lambda sender, **kwargs: False if sender == Model else True)
post_save.connect(
    receiver=lambda *args, **kwargs: log_create_reverse_fk_to_opportunity(
            action=custom_action, *args, **kwargs,
    ),
    sender=model,
)

```

I'll close the PR and the issue soon. Thanks

hramezani:
Thanks @hoangquochung1110 for letting me know

CleitonDeLima:
Hello,
I was thinking, if AuditlogModelRegistry was a singleton, could this idea work for multiple instances?

PR #641, opened 2024-05-07
add actor email

hamedsh:
# add actor_email
deleting a user from the Users table will make actor_id useless (its only a bunch of characters). by adding email as static content, it can be usable forever.

### DIFF
--- a/auditlog/context.py
+++ b/auditlog/context.py
@@ -24,7 +24,11 @@ def set_actor(actor, remote_addr=None, remote_port=None):
     auditlog_value.set(context_data)
 
     # Connect signal for automatic logging
-    set_actor = partial(_set_actor, user=actor, signal_duid=context_data["signal_duid"])
+    set_actor = partial(
+        _set_actor,
+        user=actor,
+        signal_duid=context_data["signal_duid"],
+    )
     pre_save.connect(
         set_actor,
         sender=LogEntry,
@@ -62,6 +66,7 @@ def _set_actor(user, sender, instance, signal_duid, **kwargs):
             and instance.actor is None
         ):
             instance.actor = user
+            instance.actor_email = hasattr(user, "email") and user.email or None
 
         instance.remote_addr = auditlog["remote_addr"]
         instance.remote_port = auditlog["remote_port"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 #### Improvements
 
+- feat: Added 'LogEntry.actor_email` field. ([#641](https://github.com/jazzband/django-auditlog/pull/641))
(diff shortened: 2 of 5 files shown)


auditlog/context.py
```diff
@@ -23,7 +23,11 @@ def set_actor(actor, remote_addr=None):
     auditlog_value.set(context_data)
 
     # Connect signal for automatic logging
-    set_actor = partial(_set_actor, user=actor, signal_duid=context_data["signal_duid"])
+    set_actor = partial(
+        _set_actor,
+        user=actor,
+        signal_duid=context_data["signal_duid"],
+    )
```
hramezani, line 30: what does that change do?

hramezani:
Thanks @hamedsh for this PR. Please

- Add change log entry for this change
- Add some tests to show 1. email will be logged in `LogEntry` model and 2. email will remain in after user deletion

hamedsh:
> Thanks @hamedsh for this PR. Please
> 
> * Add change log entry for this change
> * Add some tests to show 1. email will be logged in `LogEntry` model and 2. email will remain in after user deletion

@hramezani can you please take a look?

auditlog/context.py
```diff
@@ -62,6 +66,7 @@ def _set_actor(user, sender, instance, signal_duid, **kwargs):
             and instance.actor is None
         ):
             instance.actor = user
+            instance.actor_email = user.email
```
hramezani, line 69: what happens if we have a custom user model without an `email` field? it has to be set to `None`

hramezani:
Thanks @hamedsh for updating the PR. I left a comment. Also, please rebase and add a changelog entry in https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md

hramezani (approved):

hramezani:
Thanks @hamedsh

auditlog/context.py
```diff
@@ -62,6 +66,7 @@ def _set_actor(user, sender, instance, signal_duid, **kwargs):
             and instance.actor is None
         ):
             instance.actor = user
+            instance.actor_email = hasattr(user, "email") and user.email or None
```
GreatBahram, line 69: :thinking: this is a bit cryptic, why not:

`getattr(user, "email", None)`

hramezani: feel free to riase a PR and change it

csaska:
> add actor_email
deleting a user from the Users table will make actor_id useless (its only a bunch of characters). 

I would expect most apps to not delete users and instead to utilize `User.is_active`. Deleting _any object_ is destructive and makes the audit log hard to interrogate and reason about. 

> by adding email as static content, it can be usable forever.

While true, email is very frequently a PII concern and thus something a majority of apps may not want to persist to an auditlog.

PR #643, opened 2024-05-15
Disable admin interface from settings.py

gferrate:
Disables admin interface by adding `AUDITLOG_DISABLE_ADMIN_INTERFACE = False` in settings.py

### DIFF
--- a/auditlog/admin.py
+++ b/auditlog/admin.py
@@ -1,5 +1,6 @@
 from functools import cached_property
 
+from django.conf import settings
 from django.contrib import admin
 from django.contrib.auth import get_user_model
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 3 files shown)


auditlog/admin.py
```diff
@@ -57,3 +57,7 @@ def has_delete_permission(self, request, obj=None):
     def get_queryset(self, request):
         self.request = request
         return super().get_queryset(request=request)
+
+
+if not settings.AUDITLOG_DISABLE_ADMIN_INTERFACE:
```
hramezani, line 62: please change the settings name to `AUDITLOG_ENABLE_ADMIN_INTERFACE` with default value of `True`

hramezani:
Thanks @gferrate for this PR. I left some comments.

Please add a test and also changelog into https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md

auditlog/admin.py
gferrate: Sure! Makes sense

docs/source/usage.rst
```diff
@@ -463,3 +463,10 @@ Django Admin integration
 
 When ``auditlog`` is added to your ``INSTALLED_APPS`` setting a customized admin class is active providing an enhanced
 Django Admin interface for log entries.
+
+
```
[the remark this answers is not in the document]
gferrate: 👍

PR #648, opened 2024-05-24
DecimalField trailing zeros fix

ganiserb:
Working on a fix for this issue.

### DIFF
--- a/auditlog_tests/models.py
+++ b/auditlog_tests/models.py
@@ -200,6 +200,16 @@ class DateTimeFieldModel(models.Model):
     history = AuditlogHistoryField(delete_related=True)
 
 
+class DecimalFieldModel(models.Model):
+    """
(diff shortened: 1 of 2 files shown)


ganiserb:
I am closing this PR. I was blaming django-auditlog for the problem that you can see here reproducible by the tests. But it appears that the problem is in django itself. It seems that django is changing the Decimal value it stores when you save again.

CleitonDeLima:
> I am closing this PR. I was blaming django-auditlog for the problem that you can see here reproducible by the tests. But it appears that the problem is in django itself. It seems that django is changing the Decimal value it stores when you save again.

I don't know if it's a django problem, when saving an integer value, for example, in the database it is saved with decimal places.
Tavelz, the auditlog should treat decimal values ​​and not differentiate between equal values, 0 should be equal to 0.00, I believe.

PR #650, opened 2024-06-03
Use `attname` to get pk value

evanandrews-xrd:
If a model's primary key is another model, the current implementation of `LogEntryManager._get_pk_value` will needlessly pull in the entire object only to extract its primary key and discard it. This change gets the primary key directly from the first object.

The `attname` field isn't particularly well documented in Django, but it is briefly mentioned [here](https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.Field.pre_save).

On the base `Field` class, it is the same as `name`:

https://github.com/django/django/blob/adae619426b6f50046b3daaa744db52989c9d6db/django/db/models/fields/__init__.py#L973-L974

Whereas for `ForeignKey`, it is `{name}_id`:

https://github.com/django/django/blob/adae619426b6f50046b3daaa744db52989c9d6db/django/db/models/fields/related.py#L1117-L1118

Happy to remove the `assert` if you'd prefer not to use it.

### DIFF
--- a/auditlog/models.py
+++ b/auditlog/models.py
@@ -213,12 +213,11 @@ class LogEntryManager(models.Manager):
         :type instance: Model
         :return: The primary key value of the given model instance.
         """
-        pk_field = instance._meta.pk.name
+        pk_field = instance._meta.pk.attname
         pk = getattr(instance, pk_field, None)
 
         # Check to make sure that we got a pk not a model object.
-        if isinstance(pk, models.Model):
-            pk = self._get_pk_value(pk)
+        assert not isinstance(pk, models.Model)
         return pk
 
     def _get_serialized_data_or_none(self, instance):
--- a/auditlog/models.py
+++ b/auditlog/models.py
@@ -213,10 +213,12 @@ class LogEntryManager(models.Manager):
         :type instance: Model
         :return: The primary key value of the given model instance.
         """
+        # Should be equivalent to `instance.pk`.
         pk_field = instance._meta.pk.attname
         pk = getattr(instance, pk_field, None)
 
         # Check to make sure that we got a pk not a model object.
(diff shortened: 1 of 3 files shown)


evanandrews-xrd:
I've added some tests, although looking at this further it might not be necessary to have `LogEntryManager._get_pk_value` at all. Django provides the `Model.pk` property which gives you the value of the primary key:

https://docs.djangoproject.com/en/5.0/ref/models/instances/#the-pk-property

https://github.com/django/django/blob/adae619426b6f50046b3daaa744db52989c9d6db/django/db/models/base.py#L653-L663

The only difference now between `LogEntryManager._get_pk_value(instance)` and `instance.pk` is that the former will return `None` if the primary key attribute is (somehow) missing, while the latter will raise `AttributeError`.

If you'd prefer, we can probably just delete the method and use `instance.pk` where it's called.

hramezani:
Thanks @evanandrews-xrd for the update and providing more information.

> The only difference now between LogEntryManager._get_pk_value(instance) and instance.pk is that the former will return None if the primary key attribute is (somehow) missing, while the latter will raise AttributeError.

> If you'd prefer, we can probably just delete the method and use instance.pk where it's called.

Then we need to catch `AttributeError` everywhere we use `_get_pk_value` to prevent breaking change. 
I would prefer to keep the method

hramezani (dismissed):

hramezani:
Let me know if you are done with this PR. then we can merge it

evanandrews-xrd:
I've just added a couple of extra comments to include what I touched on above. Happy for you to merge when ready. Thanks!

hramezani (approved):

hramezani:
Thanks @evanandrews-xrd

PR #652, opened 2024-06-10
Reorder condition in get_field_value function

GreatBahram:
Hello there,

Some projects have their own fields that are not connected to any model. With the previous approach, the code would have broken, but with this order, it will consider those too. What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - Fixed a problem when setting `Value(None)` in `JSONField` ([#646](https://github.com/jazzband/django-auditlog/pull/646))
 - Fixed a problem when setting `django.db.models.functions.Now()` in `DateTimeField` ([#635](https://github.com/jazzband/django-auditlog/pull/635))
-- Changed the order of `rel_class` and `one_to_one` and `many_to_one` in `get_field_value` function. ([#XXX](https://github.com/jazzband/django-auditlog/pull/652))
+- Changed the order of `rel_class` and `one_to_one` and `many_to_one` in `get_field_value` function. ([#652](https://github.com/jazzband/django-auditlog/pull/652))
 
 ## 3.0.0 (2024-04-12)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fixed a problem when setting `Value(None)` in `JSONField` ([#646](https://github.com/jazzband/django-auditlog/pull/646))
 - Fixed a problem when setting `django.db.models.functions.Now()` in `DateTimeField` ([#635](https://github.com/jazzband/django-auditlog/pull/635))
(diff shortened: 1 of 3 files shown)


hramezani:
Thanks @GreatBahram for this PR.
Could you please:

- Add a test to show what problem you are going to fix
- Add an entry to the changelog file.

GreatBahram:
> Thanks @GreatBahram for this PR. Could you please:
> 
>     * Add a test to show what problem you are going to fix
> 
>     * Add an entry to the changelog file.


Hi there, I thought it would be beneficial to check the `rel_class` attribute first and then validate the `one_to_one` and `many_to_one` attributes.

[This change](https://github.com/jazzband/django-auditlog/commit/c6496292259aab1363d4acd7ff6014d9acd470eb) was disrupting our codebase. We had several non-database fields that functioned correctly before updating the auditlog. The workaround for us was to set the `many_to_one` and `one_to_one` attributes to make sure we don't end up in this newly introduced condition. I wonder if it makes sense to do it this way. what do you think?

hramezani:
> GreatBahram

Agree, that make sense to check the `rel_class` before. we were not aware of your usecase. that's why I asked for a test case. so, next time if somebody change the condition order, the test will fail.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Fixed a problem when setting `Value(None)` in `JSONField` ([#646](https://github.com/jazzband/django-auditlog/pull/646))
 - Fixed a problem when setting `django.db.models.functions.Now()` in `DateTimeField` ([#635](https://github.com/jazzband/django-auditlog/pull/635))
+- Changed the order of `rel_class` and `one_to_one` and `many_to_one` in `get_field_value` function. ([#XXX](https://github.com/jazzband/django-auditlog/pull/652))
```
hramezani, line 9:
```suggestion
- Changed the order of `rel_class` and `one_to_one` and `many_to_one` in `get_field_value` function. ([#652](https://github.com/jazzband/django-auditlog/pull/652))
```

auditlog_tests/tests.py
```diff
@@ -2164,6 +2164,19 @@ def test_log_entry_created_if_obj_strings_are_same_for_two_objs(self):
         self.assertEqual(int(log_create.changes_dict["related"][1]), one_simple.id)
         self.assertEqual(int(log_update.changes_dict["related"][1]), two_simple.id)
 
+    def test_rel_class_checked_first(self):
```
hramezani, line 2167: Better to create a model as you described in the PR description and test with a real model instead of mocking

GreatBahram:
Hi there, while working on the test case and investigating Django non-database fields reached this conclusion we should have set `many_to_many`, `many_to_one` and `one_to_one` to `None`, so I'll close this issue, thanks for the support :sunflower:

PR #653, opened 2024-06-12
Sync django query and postgres query

GreatBahram:
The [upgrade documentation](https://django-auditlog.readthedocs.io/en/latest/upgrade.html) recommends a multi-stage upgrade, during which some records might have a `changes` field that is no longer empty. This case is handled in the Django query but not in the Postgres one.

Additionally, if there are many records and the upgrade fails, we are returned to the initial state, which is not ideal.

This change should take care of these problems :crossed_fingers:


hramezani (approved):

hramezani:
Thanks @GreatBahram

PR #671, opened 2024-10-01
Added remote port

### LINKED ISSUE #668: Create remote_port field
I wonder if it wouldn't be interesting to keep a record of the port in the logs, in some cases it might be useful to keep this information.

In the [middleware](https://github.com/jazzband/django-auditlog/blob/07a3a83828a03ff4479bafe632dc8921f7a87d6d/auditlog/middleware.py#L31) it is ignored.

CleitonDeLima:
Resolve #668 

If there is another header to get the port I can include it too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,4 @@ venv.bak/
 
 ### JetBrains
 .idea/
+.vscode/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Next Release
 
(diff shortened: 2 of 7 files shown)


hramezani:
Thanks @CleitonDeLima for this PR. please add a new entry in changelog https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md in `Improvements` section

hramezani (approved):

hramezani:
Thanks @CleitonDeLima

PR #681, opened 2024-10-20
feat: Add truncate option to flush command

### LINKED ISSUE #680: Add new option to truncate log entries when calling flush command
I wonder if it is a good idea to truncate (and possibly restart the identity) for the log entry table instead of deleting records. This can be particularly useful in a scenario where the log entry table contains many records.

Obviously, Django ORM has no truncate method, and depending on the database vendor, a raw SQL query should be executed using a connection cursor.

BinDruid:
Fixes #680 
Passing `--truncate` to flush command executes a truncate SQL statement, instead of deleting log records.

### DIFF
--- a/auditlog/management/commands/auditlogflush.py
+++ b/auditlog/management/commands/auditlogflush.py
@@ -1,6 +1,7 @@
 import datetime
 
 from django.core.management.base import BaseCommand
+from django.db import connection
 
 from auditlog.models import LogEntry
 
@@ -25,11 +26,24 @@ class Command(BaseCommand):
             dest="before_date",
             type=datetime.date.fromisoformat,
         )
+        parser.add_argument(
+            "-t",
+            "--truncate",
+            action="store_true",
+            default=None,
+            help="Truncate log entry table.",
+            dest="truncate",
+        )
 
     def handle(self, *args, **options):
         answer = options["yes"]
+        truncate = options["truncate"]
         before = options["before_date"]
-
+        if truncate and before:
+            self.stdout.write(
+                "Truncate deletes all log entries and can not be passed with before-date."
+            )
+            return
         if answer is None:
             warning_message = (
                 "This action will clear all log entries from the database."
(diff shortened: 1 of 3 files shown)


auditlog/management/commands/auditlogflush.py
```diff
@@ -42,11 +56,45 @@ def handle(self, *args, **options):
...
+            self.stdout.write("Aborted.")
+            return
+
+        if not truncate:
             entries = LogEntry.objects.all()
             if before is not None:
                 entries = entries.filter(timestamp__date__lt=before)
             count, _ = entries.delete()
             self.stdout.write("Deleted %d objects." % count)
+            return
```
hramezani, line 69: I think this line here is useless because we don't have any code after the `if - else` block.

auditlog/management/commands/auditlogflush.py
```diff
@@ -42,11 +56,45 @@ def handle(self, *args, **options):
...
+    def __init__(self, database_vendor: str, table_name: str):
+        self.database_vendor = database_vendor
+        self.table_name = table_name
+
+    @property
+    def is_not_supported(self):
+        return self.database_vendor not in self.SUPPORTED_VENDORS
+
+    def to_sql(self):
+        return f"TRUNCATE TABLE {self.table_name};"
```
hramezani, line 100: I think the class could be simple. I would suggest a class without a constructor and only two static methods:

- `support_truncate_statement` -> accepts `database_vendor`
- `to_sql` -> accepts `table_name`

hramezani:
Thanks @BinDruid for this PR. I left some comments. Please address the coverage as well.
Then we are good to merge it.

auditlog/management/commands/auditlogflush.py
BinDruid: Good point, I refactored this class.

auditlog/management/commands/auditlogflush.py
BinDruid: It got fixed, thanks for suggestion.

BinDruid:
@hramezani 
I refactored the `TruncateQuery` class and added another test case to cover unsupported databases.

hramezani:
Thanks @BinDruid for the update.

Please add a change log entry in (CHANGELOG.md)[https://github.com/jazzband/django-auditlog/blob/master/CHANGELOG.md].

Sorry, I forgot to mention it before.

hramezani (approved):

hramezani:
Thanks @BinDruid

PR #682, opened 2024-10-28
Configure actions in auditlog.register

### LINKED ISSUE #424: Enable/disable logging actions (Create, Update, Delete) on specific Models
Hi,
First of all, thank you for this excellent app. I love how simple yet powerful this is.
My favorite part is being able to configure everything in `settings.py` so each users of a django project can decide if/what they want to track.

I was wondering if there is already a way (couldn't find it in the docs) to skip logs for Creation or Deletion of objects, or to only log Update actions.
One example would be having a lot of "read-only" type objects in a project where we don't want to clog the audit log table with all the creation of objects, but simply see when one is changed.
(even though I mentioned read-only, some objects can still be edited in the django admin interface).

If it doesn't exist yet, maybe this could be implemented as a list of Action IDs that could be set on a model.

Thanks again for the great work!

CleitonDeLima:
Resolve #424


hramezani:
Thanks @CleitonDeLima for this PR.

I think this PR will introduce a breaking change. the `AuditlogModelRegistry` class is exposed and this PR will change the class constructor signature.

In the current implementation, we can have multiple instances of `AuditlogModelRegistry` with different `create, update, delete, ..` configs.
```py
from auditlog.registry import AuditlogModelRegistry, auditlog

auditlog.register(MyModel)  # registers model with all actions

custom_auditlog = AuditlogModelRegistry(delete=False, access=False)
custom_auditlog.register(MyOtherModel)
```

This will not be possible by your change and some people's code might depend on this.

Also, by moving the adding to `_signals` to `def register`, by registering a model this will change for all other models as well which is wrong.

```py
from auditlog.registry import  auditlog

auditlog.register(MyModel)  # registers model with all actions

# This will change the signals in `_signals` for the whole instance and will change the log behavior for all registered model
custom_auditlog.register(MyOtherModel, actions={'DELETE': False})
```

I think the `action` can be added to `def register` but it shouldn't change the whole `auditlog` behavior.
I would suggest handling the override actions in `_connect_signals` method.

Please remember to add more tests and change the documentation if you want to continue.

CleitonDeLima:
Hi @hramezani, thanks for reviewing

> In the current implementation, we can have multiple instances of `AuditlogModelRegistry` with different `create, update, delete, ..` configs.

Is it possible to have multiple instances? It seems to me that the application currently does not support this. The [`auditlog`](https://github.com/jazzband/django-auditlog/blob/master/auditlog/registry.py#L356) instance is used elsewhere. This problem is described here #617.


> Also, by moving the adding to `_signals` to `def register`, by registering a model this will change for all other models as well which is wrong.

This really is weird, I'll think of an alternative.


> Please remember to add more tests and change the documentation if you want to continue.

I'll see how it works to change the documentation for this, it's a new world for me :smile:.

PR #686, opened 2024-11-10
fix: use sender for m2m signal dispatch connection

### LINKED ISSUE #685: Logs duplicated when used in conjunction with django-taggit and a custom tag model shared between models
I have a project using [django-taggit](https://github.com/jazzband/django-taggit) that I'd like to add django-auditlog to.

In this project I'm using a custom tag model associated with numerous other models per [django-taggit's documentation on how to use a custom tag](https://django-taggit.readthedocs.io/en/latest/custom_tagging.html#custom-tag).

For this use case, when using the `m2m_fields` configuration, I get N logs for each model registered with the connection tot custom tag. The log is correctly associated with the model the action occurred on, it just repeated N times.

This appears to happen because signal connection is made to the same `m2m_model` multiple times here -- https://github.com/jazzband/django-auditlog/blob/d4f99c272963766e4accc2e747bb99c1e6641bc8/auditlog/registry.py#L202-L206

Repro repo: https://github.com/babybuddy/django-auditlog-685/

cdubz:
This fix adds support for a use case where a single m2m through model is used on multiple models. When the receiver is used for the dispatch uid in this use case it cause duplicated logs because the through model signal connection happens multiple times.

By changing the m2m signal connection to use the sender for the dispatch uid this duplication is prevented because the signal connection only happens once for the through model.

Fixes #685

I wasn't able to run tests as I don't have a postgres database connection handy and I couldn't get the tests to work with SQLite. Test still need to be run and probably a test should be added for this use case.


hramezani:
Thanks @cdubz for reporting the issue and for the fix PR.

You can easily run a Postgres in your machine by `docker run --name some-postgres -e POSTGRES_PASSWORD=postgres -p 5432:5432 -d postgres` and then you can add a test.

Please remember to add a changelog entry as well

cdubz:
Oh yeah! I will add a test and changelog entry. Thanks!

cdubz:
Ok, I think this is ready for review again -- not used to this set of formatting tools 😁

I confirmed the test I added reproduces the issue on `master` (and fails).

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - feat: Added `LogEntry.remote_port` field. ([#671](https://github.com/jazzband/django-auditlog/pull/671))
 - feat: Added `truncate` option to `auditlogflush` management command. ([#681](https://github.com/jazzband/django-auditlog/pull/681))
 - feat: Added `AUDITLOG_CHANGE_DISPLAY_TRUNCATE_LENGTH` settings to keep or truncate strings of `changes_display_dict` property at variable length. ([#684](https://github.com/jazzband/django-auditlog/pull/684))
+- fix: Use sender instead of receiver for `m2m_changed` signal ID to prevent duplicate entries for models that share a related model. ([#686](https://github.com/jazzband/django-auditlog/pull/686))
```
hramezani, line 10: Please move it to the section below(`Fixes`)

cdubz: Oops — fixed.

hramezani:
Thanks @cdubz

cdubz:
Thanks for the quick review!

PR #705, opened 2025-03-07
Use recommended manager to query related models

sebastianmanger:
When using custom managers, I stumbled over unexpected behavior of the `log_update` signal. The signal directly uses the `objects` manager. From https://docs.djangoproject.com/en/5.1/topics/db/managers/#default-managers:

> If you’re writing some code that must handle an unknown model, for example, in a third-party app that implements a generic view, use this manager (or [_base_manager](https://docs.djangoproject.com/en/5.1/topics/db/managers/#django.db.models.Model._base_manager)) rather than assuming the model has an objects manager.

This is the result of some simple grepping, so it might not be complete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ via `AUDITLOG_MASK_TRACKING_FIELDS` setting. ([#702](https://github.com/jazzband
 - fix: Use sender instead of receiver for `m2m_changed` signal ID to prevent duplicate entries for models that share a related model. ([#686](https://github.com/jazzband/django-auditlog/pull/686))
 - Fixed a problem when setting `Value(None)` in `JSONField` ([#646](https://github.com/jazzband/django-auditlog/pull/646))
 - Fixed a problem when setting `django.db.models.functions.Now()` in `DateTimeField` ([#635](https://github.com/jazzband/django-auditlog/pull/635))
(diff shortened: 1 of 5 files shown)


hramezani:
Thanks @sebastianmanger for this PR.

Please add test if possible and add an entry in changelog

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ via `AUDITLOG_MASK_TRACKING_FIELDS` setting. ([#702](https://github.com/jazzband
 - fix: Use sender instead of receiver for `m2m_changed` signal ID to prevent duplicate entries for models that share a related model. ([#686](https://github.com/jazzband/django-auditlog/pull/686))
 - Fixed a problem when setting `Value(None)` in `JSONField` ([#646](https://github.com/jazzband/django-auditlog/pull/646))
 - Fixed a problem when setting `django.db.models.functions.Now()` in `DateTimeField` ([#635](https://github.com/jazzband/django-auditlog/pull/635))
+- Use the [default manager](https://docs.djangoproject.com/en/5.1/topics/db/managers/#default-managers) instead of `objects` to support custom model managers.
```
hramezani, line 22: Please follow the change log format. 
```suggestion
- Use the [default manager](https://docs.djangoproject.com/en/5.1/topics/db/managers/#default-managers) instead of `objects` to support custom model managers. ([#705](https://github.com/jazzband/django-auditlog/pull/705))
```

sebastianmanger: Done, thanks!

hramezani (approved):

hramezani:
Thanks @sebastianmanger

PR #706, opened 2025-03-11
Let CI also check for missing migrations

sebastianmanger:
This was discovered during development of #705 

The field `actor_email` of the `LogEntry` model [does not define a verbose name](https://github.com/jazzband/django-auditlog/blob/master/auditlog/models.py#L387), but the [migration does](https://github.com/jazzband/django-auditlog/blob/master/auditlog/migrations/0017_add_actor_email.py).

### Fix

The actual fix is [very small](https://github.com/jazzband/django-auditlog/pull/706/commits/5d3460d64d3d66cc2250ff6e5baa866624d12cc3#diff-9ec8d9f45b533f800f42a5860231256588514d2af3a7c0dea41223c64459a19cR387). I could also provide an additional migration file and leave the model as is - but chose to update the model, as it seems more consistent with the other fields. 

### Checking migrations

As I find it very useful to check missing migrations during CI runs, I [added the changes necessary](https://github.com/jazzband/django-auditlog/pull/706/commits/6a21ccb5e0eef95bfb497c19998d79e90b444c6d) to do so in this project. I realize that this is a large-ish commit with some design decision and would be happy to drop it if it's considered "too much".

An alternative approach would be to simply copy the whole `auditlog_tests` folder for the migration check: 

```
[testenv:py39-checkmigrations]
description = Check for missing migrations
deps =
    Django>=4.2
    psycopg2
setenv =
    DJANGO_SETTINGS_MODULE = auditlog_tests.test_settings
commands =
    # The test_settings in the folder "auditlog_tests" includes "auditlog_tests" as module, so duplicate it.
    cp -R auditlog_tests auditlog_tests
    python auditlog_tests/manage.py makemigrations --check --dry-run
    rm -rf ./auditlog_tests/auditlog_tests
allowlist_externals = cp,rm
# Copied folder must also be cleaned up on error
ignore_errors = True
```

I did not add a Changelog entry, as the fix affects unreleased changes - and the CI stuff is "internal" only.

### Questions

Could you please let me know if I should update this PR so that:
* Only the model is fixed (no additional tox checks)?
* If you like the tox-check, do you prefer the one that refactors the project setup, or the "copy folder" approach?


hramezani:
Thanks @sebastianmanger

PR #725, opened 2025-05-31
Add support for custom masking functions

### LINKED ISSUE #723: Add support for user-defined masking functions (mask_callable) in addition to mask_fields
I’m using `mask_fields,` but it only masks the first half of the string.
so I’d like to propose a new feature

currently, `mask_fields` only masks the first half of a string. To enable full masking, regex-based masking, and other custom policies, we want to support user-defined masking functions.

### Proposal 
- Configure via a dotted-path string (e.g., `"custom_app.utils.full_mask"`).  
- If `mask_callable` is not provided, fall back to the default mask function, preserving the existing `mask_fields` behavior for backward compatibility.

### Benefits
- Flexible masking strategies: full, pattern-based(regex...), or any custom logic.   

any feedback is welcome!

2ykwang:
Resolve #723

### DIFF
--- a/auditlog/conf.py
+++ b/auditlog/conf.py
@@ -60,3 +60,7 @@ settings.AUDITLOG_CHANGE_DISPLAY_TRUNCATE_LENGTH = getattr(
 settings.AUDITLOG_STORE_JSON_CHANGES = getattr(
     settings, "AUDITLOG_STORE_JSON_CHANGES", False
 )
+
+settings.AUDITLOG_DEFAULT_MASK_CALLABLE = getattr(
+    settings, "AUDITLOG_DEFAULT_MASK_CALLABLE", None
+)
--- a/auditlog/diff.py
+++ b/auditlog/diff.py
@@ -1,12 +1,13 @@
 import json
 from datetime import timezone
-from typing import Optional
+from typing import Callable, Optional
 
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import NOT_PROVIDED, DateTimeField, ForeignKey, JSONField, Model
 from django.utils import timezone as django_timezone
 from django.utils.encoding import smart_str
+from django.utils.module_loading import import_string
 
 
 def track_field(field):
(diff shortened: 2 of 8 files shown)


auditlog/conf.py
```diff
@@ -60,3 +60,7 @@
 settings.AUDITLOG_STORE_JSON_CHANGES = getattr(
     settings, "AUDITLOG_STORE_JSON_CHANGES", False
 )
+
+settings.AUDITLOG_DEFAULT_MASK_CALLABLE = getattr(
```
hramezani, line 64: Let's call it `AUDITLOG_MASK_CALLABLE`, because the default one is `mask_str`

hramezani:
Thanks @2ykwang for this PR.

Some failing tests are not related to this PR. We need to address them before merging this PR. Please create a separate PR if you are interested in fixing them.

Also, pelase add a changelog entry.

auditlog/conf.py
2ykwang: thank you! fixed. - 7b66442 @hramezani

hramezani:
Thanks @2ykwang for the update. LGTM 👍 

There are CI failure that are not related to this PR but have to be fixed before merging this PR

2ykwang:
@hramezani 
> Thanks @2ykwang for the update. LGTM 👍

thank you!

> There are CI failure that are not related to this PR but have to be fixed before merging this PR

 it seems the failures are occurring in the django 6.0 (djangomain branch (unstable)) tests
we need to fix that?

hramezani:
> @hramezani
> 
> > Thanks @2ykwang for the update. LGTM 👍
> 
> thank you!
> 
> > There are CI failure that are not related to this PR but have to be fixed before merging this PR
> 
> it seems the failures are occurring in the django 6.0 (djangomain branch (unstable)) tests we need to fix that?

Yes, it is from Django main branch. I think we need to fix the failing tests. Probably we need to check Django version in the test and change the assertion fo Django >= 6

2ykwang:
> Yes, it is from Django main branch. I think we need to fix the failing tests. Probably we need to check Django version in the test and change the assertion fo Django >= 6

do you mean we should add a version check inside the test case and assert a different expected value for django >= 6.0? @hramezani

hramezani:
> > Yes, it is from Django main branch. I think we need to fix the failing tests. Probably we need to check Django version in the test and change the assertion fo Django >= 6
> 
> do you mean we should add a version check inside the test case and assert a different expected value for django >= 6.0? @hramezani

I haven't checked the change in Django. However, I think we can resolve the CI issue like this.
It would be great if you check the change in Django as well

2ykwang:
I've opened another PR. 
- #726 

please review it when you have time. @hramezani

hramezani:
> I've opened another PR.
> 
> * [Fix `get_field_value` field default value handling for Django 6.0 compatibility #726](https://github.com/jazzband/django-auditlog/pull/726)
> 
> please review it when you have time. @hramezani

Thanks. I merged your PR. please rebase!

hramezani (approved):

hramezani:
Thanks @2ykwang

2ykwang:
when is this change expected to be released? @hramezani

hramezani:
> when is this change expected to be released? @hramezani

I don't have an immediate plan. But let me know if it takes too long.
What about next month? Is it too late?

2ykwang:
would the end of this month work for you for the release? @hramezani

hramezani:
> would the end of this month work for you for the release? @hramezani

Let's do it now. Here is the release PR. please review
https://github.com/jazzband/django-auditlog/pull/729

PR #726, opened 2025-06-06
Fix `get_field_value` field default value handling for Django 6.0 compatibility

2ykwang:
related: https://github.com/jazzband/django-auditlog/pull/725#issuecomment-2943030068


hramezani:
Thanks @2ykwang

PR #737, opened 2025-07-15
added fixes for changes_display_dict when AUDITLOG_STORE_JSON_CHANGES is True

The-Alchemist:


auditlog/models.py
```diff
@@ -455,9 +455,19 @@ def changes_display_dict(self):
         if auditlog.contains(model._meta.model):
             model_fields = auditlog.get_model_fields(model._meta.model)
 
+        if settings.AUDITLOG_STORE_JSON_CHANGES:
+            changes_dict = {}
+            for field_name, values in self.changes_dict.items():
+                values_as_strings = [
+                    str(v) if v is not None else "None" for v in values
```
hramezani, line 462: can't we use `str(v) for v in values` because `str(None)='None'`

The-Alchemist: Good call.   I was getting some kind of error earlier, can't remember what it was, but all the tests still pass.  Pushed the change.

PR #741, opened 2025-07-23
Add sample project for django-auditlog testing

2ykwang:
Add  a sample Django project for testing auditlog functionality.
 
Makes it easier for developers to quickly test auditlog behavior and reproduce issues.

(based on my testing setup - feedback appreciated!)


hramezani:
Thanks @2ykwang for this PR.

Not sure we want to include this project in the code base because we need to take care of the code. I'm afraid the code becomes out of date at some point because we don't have a test for the code.

2ykwang:
Thanks feedback! @hramezani 

is the demo project unnecessary, or would you consider merging it if we establish a maintenance plan? Including a demo helps users and maintainers verify features quickly, and many popular OSS projects use a top-level examples/ or tutorial/ directory so with a clear update strategy, it should be fine.

hramezani:
> Thanks feedback! @hramezani
> 
> is the demo project unnecessary, or would you consider merging it if we establish a maintenance plan? Including a demo helps users and maintainers verify features quickly, and many popular OSS projects use a top-level examples/ or tutorial/ directory so with a clear update strategy, it should be fine.

Having the demo project is good. We need to keep the project updated and make sure that it doesn't break. So, I am mostly worry about the maintenance

2ykwang:
That makes sense. I’ll this PR to a draft for now and I’ll review the maintenance plan for the demo project and check if it’s really necessary, then update it again! 

@hramezani

PR #746, opened 2025-08-07
fixed postgre type mismatch error in mixins.py

### LINKED ISSUE #740: django.db.utils.ProgrammingError: operator does not exist: character varying = bigint
venv/lib/python3.12/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.UndefinedFunction: operator does not exist: character varying = bigint
LINE 1: ...type_id" = 17 AND "auditlog_logentry"."object_pk" IN (SELECT...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.

*In django-auditlog*, the object_pk field in LogEntry is a string field (CharField) — because the primary key of any model might be a UUID, integer, or string.

But models that use an Integer or BigInt PK, so when you use it, it causes a type mismatch: the library compares CharField = BigInt, which Postgres doesn’t allow.

Tosinibikunle:
resolves #740

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Next Release
 
+## (2025-08-14)
+
+### Fixes
+
+- fix: Ensure `object_pk` lookup works for non-string primary keys by casting PK to    string in `get_for_object()` and related queries.
+- fix: Updated tests to correctly match log entries for integer, UUID, or string  primary keys.  ([#746](https://github.com/jazzband/django-auditlog/pull/746))
+
 #### Improvements
 
 - feat: Add audit log history view to Django Admin. ([#743](https://github.com/jazzband/django-auditlog/pull/743))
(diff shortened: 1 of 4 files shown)


2ykwang (commented):
Thanks for the PR! @Tosinibikunle 

I’m currently a bit busy with work, so I’ll be able to do a full review this weekend.

In the meantime, could you please:

- Confirm whether the mismatch error occurs in `get_for_objects` (from the issue description, it seems to be triggered in the “IN query”)
- Check if the same issue also happens in `get_for_object`
- Add a test case that covers this scenario

hramezani:
Also add a changelog entry

hramezani:
@2ykwang could you review it again?

hramezani:
@2ykwang can you review this PR? probably we can include this PR in the next release.

2ykwang (changes requested):

PR #751, opened 2025-08-20
fix: defer field deletion error in auditlog

### LINKED ISSUE #709: DoesNotExist error occurs during deletion if the defer field is present
When `serialize_data=True`, a `DoesNotExist` error occurs during deletion if the `defer` field is present.

This happens because, with `serialize_data=True`, it tries to retrieve all fields not listed in `serialize_kwargs['fields']`, but the record has already been deleted.

https://github.com/jazzband/django-auditlog/blob/6414b7aedb71828469ae8aa2cf837f1ff430d65f/auditlog/models.py#L268

It should only attempt to retrieve fields that are explicitly listed in `serialize_kwargs['fields']`.

Tosinibikunle:
fixes #709

### DIFF
--- a/auditlog/models.py
+++ b/auditlog/models.py
@@ -240,7 +240,19 @@ class LogEntryManager(models.Manager):
                 "fields", self._get_applicable_model_fields(instance, model_fields)
             )
 
+    #  safeguard: if fields explicitly set → only use them
+        if "fields" in kwargs:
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks @Tosinibikunle for the PR.

Could you please rebase, fix CI errors and add a changelog entry?

Issue #542, opened 2023-07-06
Translations


Rigo-Villalta:
Hello. I see that all in AuditLog is ready for Internationalization but there is no translations.

There is a reason for that or is just that the team has no time for this. If the last is the case I would like to contribute.

joepreludian:
I would like to join as well for Brazilian Portuguese. =] Great thoughts, @Rigo-Villalta

hramezani:
Hello,

That's Great!

wlisesrivas:
I would like to contribute for the Spanish as well.

hramezani:
The base support added in https://github.com/jazzband/django-auditlog/commit/65ebec6663b869a38b7974b360ab0a2174b1d030

hramezani:
Please create PRs if you are interested in translation

hramezani:
@2ykwang can we close the issue?

2ykwang:
no problem. thanks! @hramezani

PR #762, opened 2025-09-27
Add I18N Support

2ykwang:
related issue:  #542

**support 22 languages (using LLM)**
Arabic, Bulgarian, Catalan, Czech, German, Spanish, Argentine Spanish, 
Persian, Finnish, French, Hebrew, Italian, Japanese, Kazakh, Korean, 
Dutch, Polish, Portuguese (Brazil), Russian, Slovak, Turkish, Chinese (Simplified)

- Added `.mo` files (removed from `.gitignore`)
- Updated hardcoded strings to use gettext (`_("")`)
- Added `justfile` to support i18n update workflow


hramezani:
Thanks @2ykwang for the PR. 

How did you translate the keywords? did you use any AI?

2ykwang:
@hramezani yes, I translated Korean and Japanese manually and used an LLM for the others. 
the keywords are simple, so I see this as a base for the community to refine.

2ykwang:
@hramezani I’ve kept only Japanese and Korean for now, and removed the other languages to encourage community contributions.
also added a just command to generate the base `.po` files.

justfile
```diff
@@ -0,0 +1,38 @@
+# Default recipe that lists all available commands
```
hramezani, line 1: Why didn't you use `Makefile`? Do we need to install `just`?

2ykwang: @hramezani `Justfile` isn’t as much of a “standard” as `Makefile`, but it’s now fairly common in open-source projects as a task runner. 
for `django-auditlog` porject needs (i18n, test, install), just’s command-runner model felt like a better fit. 
Its simpler syntax also makes commands easier to maintain, and it works consistently across Windows, macOS, and Linux.

2ykwang: >  also we need to install just to use it. I prefer Makefile because it is common and can cover our use case.

@hramezani makes sense. I picked Justfile since I had seen and used it frequently in past projects, but I agree Makefile is the right fit here since it’s dependency-free and already the Jazzband convention.

hramezani (approved):

hramezani:
Thanks @2ykwang

PR #763, opened 2025-10-01
Fix None type mismatch in change detection

### LINKED ISSUE #750: Wrong calculation of changes when creating a new record
I think there is a bug due to the recently introduced setting `AUDITLOG_STORE_JSON_CHANGES` and the way it works. When **creating a new record**, the `changes` field gets populated by fields that did not change.

### Environment

Python: 3.12
django-auditlog: 3.2.1

### Model

```python
class TestModel(Model):
    text = models.CharField(default="abc")
    time = models.TimeField(null=True, blank=True)
```

### Steps

1. Set `AUDITLOG_STORE_JSON_CHANGES` to `False`.
2. Create and save a new instance of the model with `text="abc", time=None`.
3. `auditlog.receivers.log_create` calls `auditlog.receivers._create_log_entry`.
4. `auditlog.diff.model_instance_diff` gets called with `old` being `None` and `new` being the new model instance. `use_json_for_changes` is `False`.
5. When `old_value` and `new_value` get calculated for the field `time`, their values are `None` and `"None"`, which means that they do not match and the field `time` gets added to the `diff` variable as `("None", "None")`, and then later `LogEntry` gets created with `changes` containing the same.

The same problem would happen for the `DateField` but not for the `DateTimeField` which has its own handling in `auditlog.diff.get_field_value`. The handling of that method's return value is inconsistent. Sometimes `smart_str` is applied, sometimes not.

2ykwang:
closes #750

### DIFF
--- a/auditlog/diff.py
+++ b/auditlog/diff.py
@@ -73,15 +73,15 @@ def get_field_value(obj, field, use_json_for_changes=False):
         try:
             model_field = obj._meta.get_field(field.name)
             default = model_field.default
-            if default is NOT_PROVIDED:
-                return None
-
-            if callable(default):
-                return default()
-
-            return default
         except AttributeError:
-            return None
+            default = NOT_PROVIDED
+
+        if default is NOT_PROVIDED:
+            default = None
+        elif callable(default):
+            default = default()
+
+        return smart_str(default) if not use_json_for_changes else default
 
     try:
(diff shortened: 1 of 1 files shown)


hramezani (approved):

hramezani:
Thanks @2ykwang

Issue #551, opened 2023-07-25
Reverse relations are being logged


aqeelat:
One issue we came across with auditlog is that we're registering all fields including reverse-related fields. Which means that:
1. we'll be hitting the db multiple times since these fields might not be pre-fetched in the QuerySet
2. we'll be updating `instance._state.fields_cache` as a side effect, which is not a documented behavior.

I will open a PR to soon to show a possible fix for the issue.

aqeelat:
@hramezani 

Sorry I didn't get around to the pr.
The issue is that when we're calling `.get_fields()` in  [diff.py](https://github.com/jazzband/django-auditlog/blob/master/auditlog/diff.py#L51), we're getting the auto created fields.

Maybe we can exclude them when we're registering the models.

```python
for field in meta.get_fields():
    if isinstance(field, ManyToManyField):
        m2m_fields.append(field.name)
    elif field.auto_created and isinstance(field, ForeignObjectRel):
        exclude_fields.append(field.name)
```

hramezani:
@aqeelat can we have a config flag for this? Something like `exclude_reverse_relation`. The default value can be `False` to don't change the old behavior.

huix11:
e

peymanahmadifar:
Has this problem not been fixed since it was reported in 2023?

hramezani:
> Has this problem not been fixed since it was reported in 2023?

It would be great if you could help and create a PR to fix the problem. I can review the PR.

peymanahmadifar:
https://github.com/jazzband/django-auditlog/pull/768

PR #768, opened 2025-10-09
Add AUDITLOG_EXCLUDE_REVERSE_RELATIONS to avoid logging reverse relation fields

peymanahmadifar:
Fixes issue #551 — Reverse relations are being logged.

Summary:
When computing model diffs we used Model._meta.get_fields() which includes
auto-created reverse relation fields. Those fields can trigger extra DB queries
(if not prefetched) and mutate instance._state.fields_cache as a side-effect.

This PR:
- Adds a small opt-in setting `AUDITLOG_EXCLUDE_REVERSE_RELATIONS` (default False).
- When enabled, reverse relation fields (auto_created & not concrete) are filtered out
  from the fields used to calculate diffs.

Notes:
- Behaviour is opt-in to remain backward compatible.
- I added a minimal documentation entry under configuration (docs).
- I also added a (minimal) test verifying that reverse relations are excluded when the setting is enabled.

Please let me know if you'd prefer to make this opt-out instead of opt-in,
or if you'd rather that the default behavior be changed. Happy to update the PR.

### DIFF
--- a/auditlog/conf.py
+++ b/auditlog/conf.py
@@ -21,6 +21,11 @@ settings.AUDITLOG_EXCLUDE_TRACKING_FIELDS = getattr(
     settings, "AUDITLOG_EXCLUDE_TRACKING_FIELDS", ()
 )
 
+# Exclude reverse relation fields across all models
+settings.AUDITLOG_EXCLUDE_REVERSE_RELATIONS = getattr(
+    settings, "AUDITLOG_EXCLUDE_REVERSE_RELATIONS", False
+)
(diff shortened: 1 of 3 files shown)


hramezani:
Thanks @peymanahmadifar for the PR.

Please add tests and a changelog entry.

@2ykwang Please review the PR

2ykwang (changes requested):
thanks for the patch! @peymanahmadifar 
could you add test cases for this change?

PR #779, opened 2025-11-05
Add new setting to control FK change representation

### LINKED ISSUE #724: Support for string representation of foreign key changes
With PR #420, the handling of changes in foreign key relationships has been updated. Instead of storing the string representation of the related object, the primary key values are now stored. Additionally, displaying such changes now requires database queries, as reflected in the updated `LogEntry.changes_display_dict` method.

Before:
```json
{ "foreign_key_field": ["foo", "bar"] }
```

After:
```json
{ "foreign_key_field": [1, 2] }
```

While this change makes sense, it makes upgrading to v3 more challenging in my case. I manually parse change data to display it to users. Additionally this new behavior introduces performance concerns. Specifically, fetching related objects for each changed foreign key field results in additional database roundtrips, which were not previously necessary.

To address this, I’d like to propose introducing a setting, such as USE_STRING_REPRESENTATION_FOR_FOREIGN_KEY_CHANGES (defaulting to False). This would allow projects like mine to retain the previous behavior and simplify the upgrade process to v3.

If this sounds acceptable, I’d be happy to prepare a PR to implement this change, as it shouldn’t be too complex.

fabianallendorf:
closes #724

This PR adds a new settings variable to control how changes to a foreign key field are stored in log entries.
The changes only affect the `get_field_value` method which skips the special handling for foreign key fields if the setting is enabled.

I decided to leave the logic of `_get_changes_display_for_fk_field` untouched, since it has a special handling for legacy Log entries anyway. Also, even if the new setting is enabled there might be log entries with primary keys. For these objects, the `changes_display_dict` works as expected.

Adds documentation for new setting and a test to verify the expected behavior. Should there be more tests?

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -469,6 +469,14 @@ After:
 
 You can use this option to enable the legacy behavior.
 
+.. warning::
+
+    This reintroduces a known issue https://github.com/jazzband/django-auditlog/issues/421
+    Commission Error: Causes unnecessary LogEntries even though no update occurrs because the string representation in memory changed
+    Omission Error: More common problem, a related object is updated to another object with the same string representation, no update is logged
+
+    Beware of these problem when enabling this setting.
+
 .. versionadded:: 3.4.0
 
 Actors


2ykwang:
Thanks @fabianallendorf for the PR.

Could you please check the failing tests?

fabianallendorf:
@2ykwang Sorry, doc tests were failing. Its fixed now

2ykwang:
thanks @fabianallendorf I'll review this by the end of this week

2ykwang:
just idea in the future, we might want to change the FK storage structure as follows

- Dict: `{"id": 123, "repr": "Object Name"}`
- Tuple:` (123, "Object Name")`

fabianallendorf:
> just idea in the future, we might want to change the FK storage structure as follows
    Dict: {"id": 123, "repr": "Object Name"}
    Tuple: (123, "Object Name")

I like this, because it gives users more flexibility. The string representation shows how the object looked at the time of the change, while storing the foreign key is a more reliable way to track updates to related objects. However, I think it should be configurable whether to use the foreign key or the string representation when displaying changes, since both options have their own advantages and disadvantages.

hramezani:
Thanks all for your great work here!

Issue #169, opened 2018-03-05
one to one relation error


ghost:
one to one relation error happening on diff.py line 80 where field object does not have attribute default if onetoonerel type field. suggested patch
            if hasattr(field, 'default'):
                value = field.default if field.default is not NOT_PROVIDED else None
            else:
                value = None

audiolion:
hey, can you provide a stack trace or test case that shows the error?

stherrien:
Same error also suggesting one line patch
value = field.default if getattr(field, 'default', NOT_PROVIDED) is not NOT_PROVIDED else None

stack trace:
File "pkgs/auditlog/diff.py", line 135, in model_instance_diff
    new_value = get_field_value(new, field)
  File "pkgs/auditlog/diff.py", line 80, in get_field_value
    value = field.default if field.default is not NOT_PROVIDED else None
AttributeError: 'OneToOneRel' object has no attribute 'default'

bynare:
I was also hitting this, when using the audit log with [django-polymorphic](https://django-polymorphic.readthedocs.io/en/stable/)

Here is a [pull request](https://github.com/jjkester/django-auditlog/pull/224) ;)

hramezani:
Fixed in https://github.com/jazzband/django-auditlog/commit/a56d0e6f784f3d0e1107f41c8c7a3f2423686c14

PR #784, opened 2025-11-24
Make diffing more robust for polymorphic models

lukasgraf:
When working with **polymorphic models**, where a child model inherits from a parent model, Django's `pre_save` signal may send model instances in a way where the `log_update()` handler receives an instance of the child as the `old` model, but an instance of the parent as the `new` model.

So, [`log_update()`](https://github.com/jazzband/django-auditlog/blob/074e6aa145476302e63657834c671cffbc046960/auditlog/receivers.py#L52-L69) will end up passing something like this to `_create_log_entry`:

```python
class Parent(models.Model):
    pass

class Child(Parent):
    pass


_create_log_entry(
    # ...
    diff_old=instance_of_child,
    diff_new=instance_of_parent,
)
```

This leads to a `FieldDoesNotExist` error when a field that only exists on the child was modified, and [`get_field_value()`](https://github.com/jazzband/django-auditlog/blob/074e6aa145476302e63657834c671cffbc046960/auditlog/diff.py#L56-L120) attempts look up that field on the parent.

<details>
 <summary>Traceback</summary>
```pytb
Traceback (most recent call last):
  # ...
  File "/usr/local/lib/python3.10/site-packages/django/db/models/base.py", line 814, in save
    self.save_base(
  File "/usr/local/lib/python3.10/site-packages/django/db/models/base.py", line 861, in save_base
    pre_save.send(
  File "/usr/local/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 176, in send
    return [
  File "/usr/local/lib/python3.10/site-packages/django/dispatch/dispatcher.py", line 177, in <listcomp>
    (receiver, receiver(signal=self, sender=sender, **named))
  File "/usr/local/lib/python3.10/site-packages/auditlog/receivers.py", line 27, in wrapper
    signal_handler(*args, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/auditlog/receivers.py", line 60, in log_update
    _create_log_entry(
  File "/usr/local/lib/python3.10/site-packages/auditlog/receivers.py", line 161, in _create_log_entry
    raise error
  File "/usr/local/lib/python3.10/site-packages/auditlog/receivers.py", line 130, in _create_log_entry
    changes = model_instance_diff(
  File "/usr/local/lib/python3.10/site-packages/auditlog/diff.py", line 251, in model_instance_diff
    new_value = get_field_value(new, field, use_json_for_changes)
  File "/usr/local/lib/python3.10/site-packages/auditlog/diff.py", line 117, in get_field_value
    return get_default_value()
  File "/usr/local/lib/python3.10/site-packages/auditlog/diff.py", line 74, in get_default_value
    model_field = obj._meta.get_field(field.name)
  File "/usr/local/lib/python3.10/site-packages/django/db/models/options.py", line 683, in get_field
    raise FieldDoesNotExist(
django.core.exceptions.FieldDoesNotExist: Parent has no field named 'parent_ptr'
```
</details>

This change makes diffing polymorphic models more robust by considering this case in `get_default_value()`. Changes to those child fields won't be tracked in these cases, but at least `django-auditlog` won't prevent the model from being saved.

---

The behavior for this case broke with the overhaul of `get_field_value()` in commit 3a58e0a9 - before, there was no call to `obj._meta.get_field(field.name)` that could raise a `FieldDoesNotExist`. After that change, `get_default_value()` may fail with that when it can't find the specified field for whatever reason.

---

There also was a somewhat similar issue #169 in the past, where django-auditlog caused problems with polymorphic models. I hope the regression test included here can help prevent breakage of these cases in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Add `AUDITLOG_USE_BASE_MANAGER` setting to override default manager use ([#766](https://github.com/jazzband/django-auditlog/pull/766))
 - Drop 'Python 3.9' support ([#773](https://github.com/jazzband/django-auditlog/pull/773))
 
+#### Fixes
+
+- Make diffing more robust for polymorphic models ([#784](https://github.com/jazzband/django-auditlog/pull/784))
+
 ## 3.3.0 (2025-09-18)
 
 #### Improvements
--- a/auditlog/diff.py
+++ b/auditlog/diff.py
@@ -3,7 +3,7 @@ from collections.abc import Callable
 from datetime import timezone
 
 from django.conf import settings
-from django.core.exceptions import ObjectDoesNotExist
+from django.core.exceptions import FieldDoesNotExist, ObjectDoesNotExist
 from django.db.models import NOT_PROVIDED, DateTimeField, ForeignKey, JSONField, Model
 from django.utils import timezone as django_timezone
 from django.utils.encoding import smart_str
@@ -74,7 +74,7 @@ def get_field_value(obj, field, use_json_for_changes=False):
         try:
             model_field = obj._meta.get_field(field.name)
             default = model_field.default
-        except AttributeError:
+        except (AttributeError, FieldDoesNotExist):
             default = NOT_PROVIDED
 
         if default is NOT_PROVIDED:


lukasgraf:
(I pushed the failing test first as a separate commit, with the intention to demonstrate that and how the test fails. The workflows needing approval first kind of threw a wrench into that idea 😝 . You can see the failure [here](https://github.com/opengisch/django-auditlog/actions/runs/19646752354/job/56263691532#step:5:126))

2ykwang:
thanks @lukasgraf for the PR! 
i checked the test results. should we also test the case where `FieldDoesNotExist` is raised?

lukasgraf:
@2ykwang thanks for the review!

The `FieldDoesNotExist` is actually raised during that test. It's a bit hard to see in the test failure since it's several nested exceptions, but it is raised [here](https://github.com/opengisch/django-auditlog/actions/runs/19646752354/job/56263691532#step:5:156).

While the `AttributeError` and `KeyError` were being handled ([here](https://github.com/jazzband/django-auditlog/blob/074e6aa145476302e63657834c671cffbc046960/auditlog/diff.py#L77C9-L77C20) and [here](https://github.com/django/django/blob/fab1f25d4ee159e256823ad427ec7074b83f0027/django/db/models/options.py#L678), respectively), the `FieldDoesNotExist` ultimately was not - which is what lead to it being raised. And which is why the only change to the code, also catching `FieldDoesNotExist`, fixes that test :)

2ykwang:
Thanks for clarifying! @lukasgraf I misunderstood. 😅

2ykwang (approved):
thank you! @lukasgraf

lukasgraf:
No worries! Thanks a ton for the quick review! 💚 🎉

hramezani:
Thanks @lukasgraf

PR #789, opened 2025-12-07
Fix AttributeError when AUDITLOG_LOGENTRY_MODEL is not set (#788)

### LINKED ISSUE #788: AttributeError: 'Settings' object has no attribute 'AUDITLOG_LOGENTRY_MODEL'
Hi,

the change introduced in 3.4.0 that allows setting a custom model does not seem to take a default value, so it fails unless you add that setting manually.

I think this is a bug and it should take the "auditlog.LogEntry" value by default

jgillard:
I hope I'm not stepping on toes with this, but I was following the issue thread and thought I'd give it a go. This is Claude Opus 4.5 assisted. I have verified that this fix works with my Django project. The Claude summary is below.

### Summary
- Fixed `AttributeError: 'Settings' object has no attribute 'AUDITLOG_LOGENTRY_MODEL'` that occurs when upgrading to 3.4.0+ without explicitly setting `AUDITLOG_LOGENTRY_MODEL`
- Added regression test to prevent this issue from recurring

### Problem
When Django's admin autodiscover imports `auditlog/admin.py` → `mixins.py`, the `get_logentry_model()` function is called at module load time before `conf.py` has a chance to apply the default value.

### Solution
Use `getattr()` with the default value directly in `get_logentry_model()`:
```python
model_string = getattr(settings, "AUDITLOG_LOGENTRY_MODEL", "auditlog.LogEntry")
```

This ensures the default is available regardless of import order, while still respecting user-defined custom values.

### Changes
- `auditlog/__init__.py`: Add `getattr()` fallback in `get_logentry_model()`
- `auditlog_tests/tests.py`: Add regression test `test_logentry_model_default_when_setting_missing`

Fixes #788

### DIFF
--- a/auditlog/__init__.py
+++ b/auditlog/__init__.py
@@ -10,10 +10,9 @@ __version__ = version("django-auditlog")
 
 
 def get_logentry_model():
+    model_string = getattr(settings, "AUDITLOG_LOGENTRY_MODEL", "auditlog.LogEntry")
     try:
-        return django_apps.get_model(
-            settings.AUDITLOG_LOGENTRY_MODEL, require_ready=False
-        )
+        return django_apps.get_model(model_string, require_ready=False)
     except ValueError:
         raise ImproperlyConfigured(
             "AUDITLOG_LOGENTRY_MODEL must be of the form 'app_label.model_name'"
@@ -21,5 +20,5 @@ def get_logentry_model():
     except LookupError:
         raise ImproperlyConfigured(
             "AUDITLOG_LOGENTRY_MODEL refers to model '%s' that has not been installed"
-            % settings.AUDITLOG_LOGENTRY_MODEL
+            % model_string
         )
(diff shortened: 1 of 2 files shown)


jgillard:
A second approach was to simply change

```py
```
to 
```py
from auditlog.conf import settings
```

inside `auditlog/__init__.py`, but I don't understand the implications of this as well, but will change to this approach that avoids the duplication of the default value if that's preferred.

jgillard:
Thanks for taking a look! I agree about the duplication. How confident are you in the alternative approach? I'm not a regular package contributor and I would hate for that change to cause someone an issue downstream. I just don't know the implications of that change, so was hoping for someone with Django package experience to ok it.

2ykwang:
@hramezani do you have any other ideas on this?

hramezani (approved):

hramezani:
Thanks @jgillard for the fix. the current approach LGTM

@2ykwang please merge the PR and make the patch release

PR #805, opened 2026-02-20
Make AUDITLOG_LOGENTRY_MODEL swappable

### LINKED ISSUE #804: Setting AUDITLOG_LOGENTRY_MODEL leaves a dangling `auditlog_logentry` table, breaking flush
```python
from django.db import models

from auditlog.models import AbstractLogEntry

class CustomLogEntryModel(AbstractLogEntry):
    role = models.CharField(max_length=100, null=True, blank=True)
```

```
DATABASE_URL=postgresql:///example uv run manage.py makemigrations polls
Migrations for 'polls':
  polls/migrations/0001_initial.py
    + Create model CustomLogEntryModel
```

After migrating and looking at my DB schema, there's both `auditlog_logentry` and `polls_customlogentrymodel`

```
+--------+----------------------------+-------+-------+
| Schema | Name                       | Type  | Owner |
|--------+----------------------------+-------+-------|
| public | auditlog_logentry          | table | jack  |
.......................................................
| public | polls_customlogentrymodel  | table | jack  |
+--------+----------------------------+-------+-------+
```

This causes issues when trying to flush the database

```
$ DATABASE_URL=postgresql:///example uv run manage.py flush --no-input
```

```
CommandError: Database example couldn't be flushed. Possible reasons:
  * The database isn't running or isn't configured correctly.
  * At least one of the expected database tables doesn't exist.
  * The SQL was invalid.
Hint: Look at the output of 'django-admin sqlflush'. That's the SQL this command wasn't able to run.
```

If we look at the output of `sqlfush` 

```
DATABASE_URL=postgresql:///example uv run manage.py sqlflush
BEGIN;
TRUNCATE "django_session", "auth_user_groups", "auth_permission", "auth_group_permissions", "auth_group", "polls_customlogentrymodel", "auth_user_user_permissions", "django_content_type", "auth_user", "django_admin_log" RESTART IDENTITY;
COMMIT;
```

Running this within `psql`

```sql
jack@(none):example> BEGIN;
 TRUNCATE "django_session", "auth_user_groups", "auth_permission", "auth_group_permissions", "auth_group", "polls_customlogentrymodel", "auth_user_user_permissions", "django_content_type", "a
 uth_user", "django_admin_log" RESTART IDENTITY;
 COMMIT;
You're about to run a destructive command.
Do you want to proceed? [y/N]: y
Your call!
BEGIN
cannot truncate a table referenced in a foreign key constraint
DETAIL:  Table "auditlog_logentry" references "django_content_type".
HINT:  Truncate table "auditlog_logentry" at the same time, or use TRUNCATE ... CASCADE.
Time: 0.004s
jack@(none):example>
```

Django rightly thinks our audit log model is `polls_customlogentrymodel`, but the fact `auditlog_logentry` exists means we can no longer flush our DB because it still has a FK constraint with `django_content_type`.

My best guess is that we need to make the log model swappable with `AUDIT_LOGENTRY_MODEL`

https://github.com/jazzband/django-auditlog/blob/3f255a02d9e261a4e6ef6bd19df41dd9922b4d1b/auditlog/migrations/0001_initial.py#L6-L77

Jackevansevo:
Fixes #804

See issue description:

Prevents a dangling `auditlog_logentry` from conflicting with the flush management command.

### DIFF
--- a/auditlog/migrations/0001_initial.py
+++ b/auditlog/migrations/0001_initial.py
@@ -71,6 +71,7 @@ class Migration(migrations.Migration):
                 "get_latest_by": "timestamp",
                 "verbose_name": "log entry",
                 "verbose_name_plural": "log entries",
+                "swappable": "AUDITLOG_LOGENTRY_MODEL",
             },
             bases=(models.Model,),
         ),


2ykwang:
Thanks @Jackevansevo for the PR! Looks like fresh installs are covered now.
do we also need to address existing users who already have an `auditlog_logentry` table?`

Jackevansevo:
> Thanks @Jackevansevo for the PR! Looks like fresh installs are covered now. do we also need to address existing users who already have an `auditlog_logentry` table?`

Is a good point.

I think our choices are either:
1. Attempt to delete / cleanup `auditlog_logentry` in apps where `AbstractLogEntry` is used
2. Do nothing

I was learning towards 2.

- Less risk / easier to review/test 

- If users haven't been hit/effected by the behaviour until now their experience doesn't change. 

- I imagine a lot of people effected would have generated their own one-off migrations to work around this anyway. We could advise/recommend this in the docs/release notes.

If I get a steer from a maintainer I'm happy to look at doing 1. and get this across the finish line.

I'd probably want some safety checks around the migration (i.e. only drop the table if there's no records present), to avoid a situation where upgrading the package inadvertently wipes data.

yaakovLowenstein:
I'm having the same issue. Any chance this will be merged?

2ykwang (approved):
@Jackevansevo  sorry for the delayed response.

I tested the scenarios below, and everything works as expected:

| Setup | Result |
|---|---|
| Existing + default model | No changes |
| New + default model | Default table is created |
| New + custom model | Only the custom table is created; `flush` succeeds |
| Existing + custom model | Existing default table remains |

I agree with 2.

Could you also add an entry to CHANGELOG.md for this change?

Thanks again for your contribution!

avin-kavish:
Thanks for the fix. Any idea when this will be shipped?

PR #817, opened 2026-04-17
Adjust docstring for `Action` class due to new `Action.ACCESS` value

der-gabe:
Viewing (i.e. accessing) objects *may* now be logged.

And higher numerical value no longer means more intrusive action, since access has the highest value and is the least intrusive.


2ykwang (approved):
Thank you! @der-gabe

